PR #2, opened 2018-07-03
Added HLS metric handling

wtao-mux:


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h
```diff
@@ -37,6 +37,16 @@ typedef NS_ENUM(NSUInteger, MUXSDKPlayerState) {
...
+    NSUInteger _lastMediaRequest;
+    NSUInteger _lastMediaRequestBytes;
+    NSUInteger _lastErrorLogEventCount;
+    NSUInteger _lastTransferEventCount;
+    double _lastTransferDuration;
+    long long _lastTransferredBytes;
+    // The following three vars are for future research only
+    // double _totalTime;
+    // long long _totalBytes;
+    // double _minBitrate;
```
jsanford8: might be worth removing these before merging, if they're not used at all.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -82,12 +82,97 @@ - (void)attachAVPlayer:(AVPlayer *)player {
...
+            AVPlayerItemAccessLogEvent *event = log.events[log.events.count - 1];
+
+            if (event.numberOfMediaRequests > 0 && event.numberOfMediaRequests != _lastMediaRequest) {
+                if (_lastTransferEventCount != log.events.count) {
+                    _lastTransferDuration= 0;
+                    _lastTransferredBytes = 0;
+                    _lastTransferEventCount = log.events.count;
+                }
+                //double transferDuration = event.transferDuration - _lastTransferDuration;
+                //long long transferredBytes = event.numberOfBytesTransferred - _lastTransferredBytes;
```
jsanford8: maybe remove these unused vars, too

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -82,12 +82,97 @@ - (void)attachAVPlayer:(AVPlayer *)player {
...
+                // !!! event.observedMinBitrate, event.observedMaxBitrate, event.observedBitrate don't seem to be accurate
+                // we did a chalres proxy dump try to calculate the bitrate, and compared with above values. It doesn't match
+                // but it matches with the calculation between line 141 to line 149
+
+                // Keep the next line of print for future reseaerch.
+                //NSLog(@"Bwm transferDuration %f, numOfBytesXferred %llu, min %fbps, max %fbps, observered %fbps", event.transferDuration, event.numberOfBytesTransferred, event.observedMinBitrate, event.observedMaxBitrate, event.observedBitrate);
+
+                NSString *domain = [self getHostName:event.URI];
+                MUXSDKBandwidthMetricData *loadData = [[MUXSDKBandwidthMetricData alloc] init];
+                loadData.requestType = @"ManifestOrMedia";
```
jsanford8: Is there any way to confirm if this includes manifests _and_ media, and maybe distinguish between the two types? That would be ideal, if possible, otherwise we may want to just omit this type (not sure if `ManifestOrMedia` is useful to our backend)

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -82,12 +82,97 @@ - (void)attachAVPlayer:(AVPlayer *)player {
...
+                loadData.requestStart = [NSNumber numberWithLong: _lastTransferDuration * 1000];
```
jsanford8: This isn't really going to be accurate, is it? I wonder... is there a way that we can get within 50 ms of accuracy, given that our timer is run every 50 ms? I'm thinking... if we use the wallclock time within the timer callback, and check when the number of requests has changed, we might have an _idea_ of when the request started, and then an idea of when the request ended.

These values work, since you're essentially patching the transferDuration into the time, but it's not correct in terms of wall-clock time, right?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -82,12 +82,97 @@ - (void)attachAVPlayer:(AVPlayer *)player {
...
+                loadData.requestHostName = (domain == nil) ? event.serverAddress : domain;
+                loadData.requestCurrentLevel = nil;
+                loadData.requestMediaStartTime = nil;
+                loadData.requestMediaDuration = nil;
+                loadData.requestVideoWidth = nil;
+                loadData.requestVideoHeight = nil;
+                loadData.requestRenditionLists = nil;
+                [self dispatchBandwidthMetric:loadData];
+
+                // manual bitrate calculation; it includes request latency, which is defined as delta between request sent and 1st byte of recsponse received.
```
jsanford8: let's remove the unused comments here, I don't think we need them, right?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -82,12 +82,97 @@ - (void)attachAVPlayer:(AVPlayer *)player {
...
+        AVPlayerItemErrorLog *error = _player.currentItem.errorLog;
+        if (error != nil && error.events.count > 0) {
+            // https://developer.apple.com/documentation/avfoundation/avplayeritemerrorlogevent?language=objc
+            if (_lastErrorLogEventCount < error.events.count) {
+                AVPlayerItemErrorLogEvent *errorEvent = error.events[error.events.count - 1];
+                NSString *domain = [self getHostName:errorEvent.URI];
+                MUXSDKBandwidthMetricData *loadData = [[MUXSDKBandwidthMetricData alloc] init];
+                loadData.requestError = errorEvent.errorDomain;
+                loadData.requestType = @"ManifestOrMedia";
+                loadData.requestUrl = errorEvent.URI;
```
jsanford8: Can we pass the whole URL in for the non-error case, too, and put the logic for getting the hostname inside Core?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -82,12 +82,97 @@ - (void)attachAVPlayer:(AVPlayer *)player {
...
+
+        AVPlayerItemErrorLog *error = _player.currentItem.errorLog;
+        if (error != nil && error.events.count > 0) {
+            // https://developer.apple.com/documentation/avfoundation/avplayeritemerrorlogevent?language=objc
+            if (_lastErrorLogEventCount < error.events.count) {
+                AVPlayerItemErrorLogEvent *errorEvent = error.events[error.events.count - 1];
+                NSString *domain = [self getHostName:errorEvent.URI];
+                MUXSDKBandwidthMetricData *loadData = [[MUXSDKBandwidthMetricData alloc] init];
+                loadData.requestError = errorEvent.errorDomain;
+                loadData.requestType = @"ManifestOrMedia";
```
jsanford8: Same comment on the type as above

wtao-mux: 'media' now

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
wtao-mux: extracting hostname requires OS dependent API (see inside the getHostName function at line 94 and line 95). i imagine the core will be transpiled so that i try to avoid OS dependent APIs as much as possible. hence, I still prefer to put the hostname setting in the integration level

jsanford8: Ahh I see that now, good call.

PR #6, opened 2018-08-01
Added Carthage support

wtao-mux:


wtao-mux:
video-demo - demos how to use a Cartfile to download dependent frameworks
carthage-archive.sh - produces a carthage release binary from update-release-frameworks build

wtao-mux:
we need to update README to remember create a Carthage zip when doing release

wtao-mux:
@jsanford8 under the apps directory, there are three demos,

1. DemoApp is an objective C iOS app using cocoapods 
2. TvDemoApp is an objective C tvOS app using cocoapods
3. video-demo is a swift iOS app using cocoapods in the before. I thought we have a DemoApp for iOS already. so I changed it to use carthage.  i can definitely duplicate it. please let me know your preference

PR #11, opened 2018-08-31
v0.1.3 restore dispatching by eventQueue

wtao-mux:


wtao-mux:
All the network code is still residing in the `MUXSDKDispatcher`. 

The fix ensures to invoke EventQueue first, then call the MUXSDKDispatcher for actual network transaction

PR #12, opened 2018-09-05
Ima

wtao-mux:


MUXSDKStats/Podfile
```diff
@@ -1,12 +1,13 @@
 # Uncomment the next line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '9.0'
```
jsanford8: Is this going to force us to iOS 9+? And if so, is that what the IMA SDK requires?

MUXSDKStats/Podfile
```diff
@@ -1,12 +1,13 @@
...
 
 target 'MUXSDKStats' do
   # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
   # use_frameworks!
 
   # Pods for MUXSDKStats
   pod 'Mux-Stats-Core', '~>2.0.0'
+  pod 'GoogleAds-IMA-iOS-SDK', '~> 3.7'
```
jsanford8: How do podfiles work? For instance, if you now load our pod, does that automatically include the IMA Pod in your project? Or is that stripped out if you don't add that yourself explicitly?

MUXSDKStats/Podfile
wtao-mux: the IMA SDK has a minimum iOS9+ requirement. that's why the uncomment

MUXSDKStats/Podfile
wtao-mux: no. framework don't embed framework. open up, the DemoApp's Podfile, you will see all three pods are mentioned

MUXSDKStats/Podfile
jsanford8: ahh interesting, okay

MUXSDKStats/Podfile
jsanford8: nice 👍

PR #18, opened 2019-08-09
only send requestStart and responseEnd data

kcking:


jsanford8:
@dylanjha - yes please, if we haven't missed the boat on that just yet. We should roll this in (can be a patch version) once we have the IMA work completed

PR #23, opened 2019-09-17
Dj/source duration bugfix

dylanjha:


dylanjha:
23-sec_2019-09-17_15-12-17

we correctly got 23 seconds here for the source duration

dylanjha:
@jsanford8 yes, I ran a new build!

PR #31, opened 2019-12-04
Add MUXSDKStats query parameter filtering

wobbals:
This change proposes a method to redact (prior known) components of a query parameter that would not be appropriate to report to an analytics service, such as access tokens to a playback URL, PII, or any derivative.

For example, if a playback URL scheme included an `accessToken` query parameter and this field needed to be redacted from analytics, integrators may invoke:

```objc
[MUXSDKStats addQueryParamFilter:@"accessToken"];
```

Prior to attaching a video player monitor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/xcshareddata/*
 **/Pods/*
 **/*.xcworkspace/*
+.DS_Store
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h
@@ -49,6 +49,8 @@ typedef NS_ENUM(NSUInteger, MUXSDKPlayerState) {
     long long _lastTransferredBytes;
 }
 
+@property (copy) NSString* videoSourceUrlOverride;
+
 - (id)initWithName:(NSString *)name andSoftware:(NSString *)software;
 - (void)attachAVPlayer:(AVPlayer *)player;
 - (void)detachAVPlayer;
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -155,6 +155,7 @@ static void *MUXSDKAVPlayerItemStatusObservationContext = &MUXSDKAVPlayerStatusO
 }
 
 - (void)dealloc {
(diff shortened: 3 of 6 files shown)


dylanjha:
@wobbals Thanks so much for opening this PR!

I think it makes sense that you (and potentially others) may want to control what sourceUrl data gets collected by Mux.

Around the specific implementation we had a short discussion about this internally. The way we're leaning right now is that something like this might be a little better.

```objc
[MUXSDKStats monitorAVPlayerViewController:player
                            withPlayerName:@"awesome"
                            playerData:playerData
                            videoData:videoData
                            videoSourceUrl:"http://yoururl.m3u8"];
```

The new pice here would be `videoSourceUrl`. If the videoSourceUrl is set at initialization time, then the sdk will always use that one and not do the auto-detecting of videoSourceUrl which it is doing now.

We would expect the vast majority of integrators to never use that param, but for cases like yours it is available. And you would be responsible for filtering the params out of the url yourself and doing whatever kind of mangling or obfuscating that you want before you pass it in.

To give you a little context, the main reasons we feel like this approach is a little better is that it would give the integrator the full control over setting the source url (if they want). I could foresee a case where filtering query parameters it not enough, an integrator might want to completely hide the source url, or obfuscate it in a different way that goes beyond query param filtering.

----

I haven't tried this yet or thought through the code-paths required for this change but I wanted to share this feedback so you know how we're thinking about it.

If you think this would work for you, feel free to take a stab at it and I'm happy to help along the way. Otherwise it's something I'm more than happy to take a look at next week.

wobbals:
Makes sense to me, thanks @dylanjha! I'll take a crack at this and see how it fits in the implementation. I didn't get too deep into the test harness for these bindings, not quite sure how I could verify the behavior in a unit test.

wobbals:
How do you feel about handling target URL changes to the player? Options I can think off the top of my head to support it:
* Ignore all changes if the videoSourceUrl input parameter is used
* Clobber videoSourceUrl if the change occurs after binding MUXSDKStats,
* Optional delegate method to mangle the URL on a per-target basis.

All of these seem clunky, not sure what the least worst option is. The first two seem to make URL overrides and multiple URLs mutually exclusive, and the third bridges the two concepts but delegates can be divisive.

dylanjha:
Closing this - resolved in Mux-Core 2.1.1.

https://github.com/muxinc/stats-sdk-objc/releases/tag/v2.1.1

Thanks for bringing this up @wobbals we appreciate your help with this SDK!

PR #33, opened 2020-01-14
bugfix in videoChangeForPlayer

dylanjha:
This was broken because:
  * we were not sending MuxCustomerPlayerData with the dispatch data event to MuxCore
  * therefore, MuxCore was ignoring requests because there was no ENV key

Solution:
  * cache the MuxCustomerPlayerData that we get for each player
  * when videoChangeForPlayer gets called, pass in the MuxCustomerPlayerData into MuxCore

I tried solving this in MuxCore, but for a few different reasons that proved v. challenging (If you want to have a bigger discussion about this I'm happy to)

---

@jsanford8 after a usual code review I would love if you could do a QA pass for me:

1. checkout master DemoApp and confirm I'm not crazy - you should see that `videoChangeForPlayer` is NOT working
1. from the master DemoApp, point to my pod like so: `pod 'Mux-Stats-AVPlayer', :git => 'https://github.com/muxinc/mux-stats-sdk-avplayer', :branch => 'dj/bugfix-video-change-for-player'`
1. See that my branch fixes the bug and `videoChangeForPlayer` now works as expected.

---


dylanjha:
I'm also considering creating this - so that a user can call either `withVideoData:` or `withVideoData:withPlayerData`, and if they call the latter, then we'll use that and update the local in-memory cache of playerData for that player.

```obj-c
+ (void)videoChangeForPlayer:(nonnull NSString *)name withVideoData:(nullable MUXSDKCustomerVideoData *)videoData withPlayerData:(nullable MUXSDKCustomerPlayerData *)playerData;
```

jsanford8:
I like the two options - fix the one that just takes video data, and create the one that can let them change the player data (if for some reason they feel the need to)

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -199,20 +205,29 @@ + (void)destroyPlayer:(NSString *)name {
...
-+ (void)videoChangeForPlayer:(nonnull NSString *)name withVideoData:(nullable MUXSDKCustomerVideoData *)videoData {
++ (void)videoChangeForPlayer:(nonnull NSString *)name withPlayerData:(nullable MUXSDKCustomerPlayerData *)playerData withVideoData:(nullable MUXSDKCustomerVideoData *)videoData {
     if (videoData) {
         MUXSDKPlayerBinding *player = [_viewControllers valueForKey:name];
         if (player) {
             [player dispatchViewEnd];
             [player dispatchViewInit];
             MUXSDKDataEvent *dataEvent = [MUXSDKDataEvent new];
             [dataEvent setCustomerVideoData:videoData];
+            MUXSDKCustomerPlayerData *playerData = [_customerPlayerDatas valueForKey:name];
```
dylanjha: remove this line - we already have playerData passed in

jsanford8:
@dylanjha - can we close this now? I believe this was fixed by the most recent PR that was merged, right?

dylanjha:
Yes!

PR #34, opened 2020-01-15
Prevent crashes when removing observers

dylanjha:
Crash report

```
Crashed: WatchProgressUpdater
0  Foundation                     0x1145aa725 _timerRelease + 10
1  MUXSDKStats                    0x1125345b1 -[MUXSDKPlayerBinding detachAVPlayer] + 202
2  MUXSDKStats                    0x112537eee +[MUXSDKStats destroyPlayer:] + 143
3  Quibi                          0x10daf2937 PlaybackQosManager.deinit + 53 (PlaybackQosManager.swift:53)
4  Quibi                          0x10daf2979 PlaybackQosManager.__deallocating_deinit + 51 (PlaybackQosManager.swift:51)
5  libswiftCore.dylib             0x11b776ed0 _swift_release_dealloc + 16

libobjc.A.dylib
objc_exception_throw
2
Foundation
-[NSObject(NSKeyValueObserverRegistration) _removeObserver:forProperty:]
3
Foundation
-[NSObject(NSKeyValueObserverRegistration) removeObserver:forKeyPath:]

Cannot remove an observer <MUXSDKAVPlayerLayerBinding 0x6000015cd930> for the key path "status" from <AVPlayerItem 0x600002cb82d0> because it is not registered as an observer.
PlaybackQosManager.deinit
keyboard_arrow_down
```

----

I think I actually see how we're getting this crash.

The `_playerItem` observer gets set up when we call `monitorAVPlayerItem`, which we call asynchronously after one of the `_player` observers fires.

Theoretically, there could be a race condition where the SDK getes initialized, but we haven't gotten an observer callback yet, and then the SDK is destroyed and we try to teardown the `_playerItem` observer.

----

This PR is a bit heavy handed - it wraps ALL attempts to remove observers - even though it's possible that our only problematic one is the one for `_playerItem`.

Some other options:

1. Is there a way to check that an observer is registered before we remove it? We can wrap all the removals in a conditional - I tried this.
1. Keep the `safelyRemovePlayerItemObserverForKeyPath` but don't change all the other observer removals. I **think** the playerItem one is the only one that is problematic. But I'm not 100% sure.
1. Check the exception message and only silence it if it matches something like `/because it is not registered as an observer/`


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -190,7 +221,7 @@ - (void)monitorAVPlayerItem {
 
 - (void)stopMonitoringAVPlayerItem {
     [MUXSDKCore destroyPlayer: _name];
-    [_playerItem removeObserver:self forKeyPath:@"status"];
```
dylanjha: This is the line that was crashing, removing an observer on an `AvPlayerItem`

dylanjha:
@jsanford8 I was able to forcibly reproduce this (see screenshot) - in this case the exception was thrown, logged and did not crash the app:

* MuxLog is part of MuxCore and it's not exposed publicly so I couldn't use that
* I decided it's still worth logging, so I used NSLog - thoughts? / should we just silence this completely?
* The exception that gets thrown is [NSRangeException](https://developer.apple.com/documentation/foundation/nsrangeexception), which is kind of generic. It's not as specific as I would like, but I think it's still worthwhile to check for.

error-removing_2020-01-15_17-03-18

jsanford8:
NSLog is fine, let's just make sure that we put a note that it's from MuxStats, and that it also specifies that this is okay/not a problem.

Yeah, checking for NSRangeException (which is a weird, weird type, but hey) makes sense to me

PR #35, opened 2020-02-03
[8100] AVPlayer SDK can miss initial `play` event at times

nidhik:
https://zube.io/muxinc/mux/c/8100

## Overview 
Makes sure that we always send events to Core in the following order :`viewinit`, `data event `containing customer player data and video data, `playerready` `play`.

## Approach
`MUXSDKStats` caches customer player data and customer video data so that the SDK has this information whenever we need to dispatch the data event in this initializing sequence to core. 

Adds a `PlayerBindingManager` that is responsible for sending the initial sequence of events. It is notified using the delegate pattern whenever the player binding is about to dispatch the `play` event so that it can send this sequence of events first, if needed. 

## Notes
Easiest to see the diff, not including the built frameworks here: https://github.com/muxinc/mux-stats-sdk-avplayer/pull/35/commits/c63e1c08049f7854a7f119dc101a7391e92b7051


Frameworks/iOS/fat/MUXSDKStats.framework/Headers/MUXSDKPlayerBinding.h
```diff
@@ -22,6 +22,9 @@ typedef NS_ENUM(NSUInteger, MUXSDKPlayerState) {
     MUXSDKPlayerStateViewEnd,
 };
 
+@protocol MUXSDKPlayDispatchDelegate
```
nidhik: Added protocol and delegate to the player binding so we have a hook right before the player binding dispatches the play event

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBindingManager.h
```diff
@@ -0,0 +1,29 @@
...
+#import "MUXSDKCustomerVideoDataStore.h"
+
+@interface MUXSDKPlayerBindingManager : NSObject<MUXSDKPlayDispatchDelegate>
+
+@property (nonatomic, weak) id<MUXSDKCustomerPlayerDataStoring> _Nullable customerPlayerDataStore;
+@property (nonatomic, weak) id<MUXSDKCustomerVideoDataStoring> _Nullable customerVideoDataStore;
+@property (nonatomic, weak) NSDictionary * _Nullable viewControllers;
+@property (nonatomic, retain) Class _Nullable core;
+
+- (void) initializeCoreForPlayer:(NSString *_Nonnull) name;
```
nidhik: Open to other/clearer names for this method. The idea is that this method is responsible for making sure MuxCore gets the initial sequence of events in the correct order

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBindingManager.m
```diff
@@ -0,0 +1,73 @@
...
+               [self dispatchDataEventForPlayerName:name playerData:playerData videoData:videoData];
+               [binding dispatchPlayerReady];
+               [self.playerReadyBindings addObject:name];
+           }
+    }
+}
+
+- (void)playerWillDispatchPlay:(NSString *) name {
+    if (![self hasInitializedPlayerBinding: name]) {
+        NSLog(@"MUXSDK-WARNING - Detected SDK initialized after playback has started.");
```
nidhik: This means the SDK is being initialized while the player is already playing. We will dispatch `ViewInitEvent`, `DataEvent`, `PlayerReadyEvent` here, then the player binding will continue and dispatch `PlayEvent`, and `PlayingEvent`.

dylanjha:
To summarize.

Happy Path:

  * implementer calls `monitorAVPlayer*`
  * we set up a `newBinding`
  * we cache the playerData and videoData
  * we set up a bindingManager
  * we assign the bindingManager as a delegate for `newBinding.playDispatchDelegate`
  * the player "plays"
  * the `rate` KVO will fire
  * we call `dispatchPlay`
  * inside `dispatchPlay`, we call `playerWillDispatchPlay` - this makes sure that core gets `viewInit` and the `data event` and `playerReady` events that it needs.

Non-happy path

  * implementer calls `monitorAVPlayer*`
  * we set up a `newBinding`
  * we cache the playerData and videoData
  * we set up a bindingManager
  * we assign the bindingManager as a delegate for `newBinding.playDispatchDelegate`
  * the `rate` KVO **does not fire** (because we're already playing)
  * we detect that the player is in the playing state and call `dispatchPlay`
  * inside `dispatchPlay`, we call `playerWillDispatchPlay` - this makes sure that core gets `viewInit` and the `data event` and `playerReady` events that it needs.

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -22,13 +24,30 @@ @implementation MUXSDKStats
 // Name => AVPlayerViewController or AVPlayerLayer
 static NSMutableDictionary *_viewControllers;
 
+static MUXSDKPlayerBindingManager *_playerBindingManager;
+static MUXSDKCustomerPlayerDataStore *_customerPlayerDatas;
+static MUXSDKCustomerVideoDataStore *_customerVideoDatas;
```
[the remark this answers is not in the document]
nidhik: @dylanjha I agree, I'll go ahead and make these changes

MUXSDKStats/MUXSDKStats/MUXSDKCustomerPlayerDataStore.m
```diff
@@ -0,0 +1,32 @@
...
+    return(self);
+}
+
+- (void) setPlayerData:(nonnull MUXSDKCustomerPlayerData *)playerData forPlayerName:(nonnull NSString *)name {
+    [self.customerPlayerDatas setValue:playerData forKey:name];
+}
+
+- (MUXSDKCustomerPlayerData *_Nullable) playerDataForPlayerName:(nonnull NSString *)name {
+    return [self.customerPlayerDatas valueForKey:name];
+}
```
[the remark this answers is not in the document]
nidhik: >MUXSDKCustomerPlayerDataStore is being used as an in-memory cache to save playerData for a particular player.

That's right.

>We were doing this before, with a dictionary in MUXSDKStats

In your branch we were using a dictionary. I was doing essentially the same thing to fix this bug, but defined a separate class and defined a protocol so that different components that need the cached metadata could reference this protocol. It also makes it easier to write tests. I do not think we are (in the current `master`) caching customer player data in the `MUXSDKStats` layer after we dispatch the data event to `MuxCore`

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBindingManager.m
nidhik: Yeah, exactly

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBindingManager.h
nidhik: One option is `newViewForPlayer:` or `startNewViewForPlayer`

nidhik:
> To summarize.
> 
> Happy Path:
> 
> * implementer calls `monitorAVPlayer*`
> * we set up a `newBinding`
> * we cache the playerData and videoData
> * we set up a bindingManager
> * we assign the bindingManager as a delegate for `newBinding.playDispatchDelegate`
> * the player "plays"
> * the `rate` KVO will fire - we call `dispatchPlay`
> * inside `dispatchPlay`, we call `playerWillDispatchPlay` - this makes sure that core gets `viewInit` and the `data event` and `playerReady` events that it needs.
> 
> Non-happy path
> 
> * implementer calls `monitorAVPlayer*`
> * we set up a `newBinding`
> * we cache the playerData and videoData
> * we set up a bindingManager
> * we assign the bindingManager as a delegate for `newBinding.playDispatchDelegate`
> * the `rate` KVO **does not fire** (because we're already playing)
> * we detect that the player is in the playing state and call `dispatchPlay`
> * inside `dispatchPlay`, we call `playerWillDispatchPlay` - this makes sure that core gets `viewInit` and the `data event` and `playerReady` events that it needs.

A few edits @dylanjha -  It's a very tricky sequence of events 😓 

Happy Path:

* implementer calls monitorAVPlayer*
* we set up a newBinding
* we cache the playerData and videoData
* we set up a bindingManager
* we assign the bindingManager as a delegate for newBinding.playDispatchDelegate
* the rate KVO does fire when we attach the player to the player binding (because it always fires when first set up) - since the rate is 0 we do not dispatch play
* we call `initializeCoreForPlayer:` on the bindingManager - this makes sure that core gets viewInit and the data event and playerReady events that it needs.
* the player "plays"
* the rate KVO will fire again - now we call `dispatchPlay` since the rate is 1
* inside `dispatchPlay`, we call `playerWillDispatchPlay`
* inside the binding manager, since we have already sent the initial sequence up to playerready, we will not do it again

Non-happy Path

* player is already playing
* implementer calls monitorAVPlayer* 
* we set up a newBinding
* we cache the playerData and videoData
* we set up a bindingManager
* we assign the bindingManager as a delegate for `newBinding.playDispatchDelegate`
* the rate KVO does fire when we attach the player to the player binding (because it always fires when first set up)
* since the rate is already 1, we detect that the player has already started playback and call `dispatchPlay`
* inside `dispatchPlay`, we call `playerWillDispatchPlay` 
* binding manager makes sure that core gets `viewInit` and the `data event` and `playerReady `events that it needs.
* we call `initializeCoreForPlayer:` on the bindingManager 
* inside the binding manager, since we have already sent the initial sequence up to playerready, we will not do it again

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBindingManager.m
```diff
@@ -0,0 +1,68 @@
...
+}
+
+
+- (BOOL) hasInitializedPlayerBinding:(NSString *) name {
+    return [self.playerReadyBindings containsObject:name];
+}
+
+- (void) initializeCoreForPlayer:(NSString *_Nonnull) name {
+    if (![self hasInitializedPlayerBinding: name]) {
+        MUXSDKPlayerBinding *binding = [self.viewControllers valueForKey:name];
```
jsanford8: well this name confused me for a minute, but then I realized it's just been this way from the beginning, and it's not necessarily a ViewController

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -199,20 +209,31 @@ + (void)destroyPlayer:(NSString *)name {
...
++ (void)videoChangeForPlayer:(nonnull NSString *)name withPlayerData:(nullable MUXSDKCustomerPlayerData *)playerData withVideoData:(nullable MUXSDKCustomerVideoData *)videoData {
     if (videoData) {
         MUXSDKPlayerBinding *player = [_viewControllers valueForKey:name];
         if (player) {
             [player dispatchViewEnd];
             [player dispatchViewInit];
             MUXSDKDataEvent *dataEvent = [MUXSDKDataEvent new];
+            [_customerVideoDataStore setVideoData:videoData forPlayerName:name];
             [dataEvent setCustomerVideoData:videoData];
+            if (playerData) {
```
jsanford8: Okay so one question here - if you call this with a null playerData, won't this have cleaned itself away and you would end up with a new view that doesn't have an environment key?

jsanford8: Or does the store that we have keep that in place on viewInit?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBindingManager.m
nidhik: Yeah, it confused me as well, we should rename this. I'll file a ticket

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
nidhik: The if `if (playerData)`  check should protect us against the `playerData` getting cleared away if someone passes in a `nil` - we only set the player data if it is not `nil`

dylanjha:
@nidhik and I just ran QA in the demoapp against this branch:

 - [x] AVPlayer happy path (initialize sdk before play)
 - [x] AVPlayer sad path (initialize sdk after play)
 - [x] AVPlayer with ads
 - [x] AVPlayer videoChange

All checks out. @jsanford8 we're going to release 1.2.0 and update documentation

MUXSDKStats/MUXSDKStatsTests/Utils/MUXSDKCore+Mock.m
```diff
@@ -0,0 +1,61 @@
...
+
+#import "MUXSDKCore+Mock.h"
+#include <objc/runtime.h>
+
+
+@implementation MUXSDKCore (Mock)
+
+static NSMutableDictionary *events;
+
++ (void) swizzleDispatchEvents {
```
dylanjha: Nice job with this. It makes it much easier for us to write tests and I can see this was the only way you were able to test the behavior that you fixed.

PR #38, opened 2020-02-13
Automatically track rendition changes.

nidhik:
## Overview
Automatically track rendition changes by listening for `AVPlayerItemNewAccessLogEntryNotification`.

## Notes

**This only includes advertised bitrate.** I would like to do automatic detection of advertised  frame rate in a separate PR. Therefore the `renditionchange` events currently always contain a `null` value for `vsofp`

## Zube
* https://zube.io/muxinc/mux/c/9807
* https://zube.io/muxinc/mux/c/9805


dylanjha:
@nidhik I'll be happy to QA this, I suppose I can do that with Charles, right?

nidhik:
@dylanjha Thanks so much! I used [Network Link Conditioner](https://nshipster.com/network-link-conditioner/) to test this but I bet you could also do it with Charles proxy. Might be good to QA it multiple ways

dylanjha:
@nidhik Ah I see, you used Network to force the automatic rendition changes were happening, that's neat. I was thinking of using Charles to verify the data in the beacons that were getting sent across the wire. Okay, I will check this out tomorrow.

nidhik:
@dylanjha Thanks! Yeah I found it was easiest to test the rendition changes with Network Link Conditioner on device and then have the device proxying through Charles to verify the beacons. I think you could also use Charles to throttle but I think Network Link Conditioner is a better simulation of poor network conditions

PR #39, opened 2020-02-14
Automatic detection of advertised  frame rate 

nidhik:
## Overview

Automatically track advertised frame rate by parsing `FRAME-RATE` attribute from the master manifest if it is present.

If we know the advertised frame rate, it will be included in the video data for playback events, including `renditionchange`.

## Zube
* https://zube.io/muxinc/mux/c/9807
* https://zube.io/muxinc/mux/c/9805


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -263,6 +290,19 @@ - (void)monitorAVPlayerItem {
                       forKeyPath:@"status"
                          options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew
                          context:MUXSDKAVPlayerItemStatusObservationContext];
+        
+        if([_player.currentItem.asset isKindOfClass:[AVURLAsset class]]) {
+            AVURLAsset *asset = (AVURLAsset *) _player.currentItem.asset;
+            NSLog(@"MUXSDK-INFO - Loading HLS master manifest from asset URL: %@", asset.URL);
+            __weak MUXSDKPlayerBinding *weakSelf = self;
```
nidhik: `__weak` here so we don't accidentally retain `self` causing a memory leak. If the player binding has been deallocated by the time the completion block is called, nothing will happen

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -250,6 +276,7 @@ - (void)detachAVPlayer {
         [_timeUpdateTimer invalidate];
         _timeUpdateTimer = nil;
     }
+    [_masterPlaylistRequest cancel];
```
nidhik: It is important we cancel any requests for the master manifest that have not yet completed. In the case of change video, we don't want the first request completing after the new video has started,

PR #40, opened 2020-02-18
Move bandwidth metric calculation from `timeUpdateTimer:` to the handler for `AVPlayerItemNewAccessLogEntryNotification`

nidhik:
## Overview

Currently, we poll to check for new entries in `AVPlayerItemAccessLog` and use new entries to calculate `MUXSDKBandwidthMetricData`. This change moves this logic to the handler for  `AVPlayerItemNewAccessLogEntryNotification`  instead of polling

Do the same for new entries in `AVPlayerItemErrorLog` - now we listen for [`AVPlayerItemNewErrorLogEntryNotification`](https://developer.apple.com/documentation/avfoundation/avplayeritemnewerrorlogentrynotification?language=objc) instead of polling. 

## Zube

- https://zube.io/muxinc/mux/c/9955


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -107,9 +104,11 @@ -(NSString *)getHostName:(NSString *)urlString {
 # pragma mark AVPlayerItemAccessLog
 
 - (void)handleAVPlayerAccess:(NSNotification *)notif {
-    AVPlayerItemAccessLog *accessLog = [((AVPlayerItem *)notif.object) accessLog];
-    [self handleRenditionChangeInAccessLog:accessLog];
-    [self calculateBandwidthMetricFromAccessLog:accessLog];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
nidhik: We should make sure to handle these notifications on the main queue:

>This notification may be posted on a different thread than the one on which the observer was registered.
https://developer.apple.com/documentation/avfoundation/avplayeritemnewaccesslogentrynotification?language=objc

nidhik:
@dylanjha @jsanford8 I wasn't able to figure out a way to do a unit test for this, because the `AVPlayerItemNewErrorLogEntryNotification` and `AVPlayerItemNewAccessLogEntryNotification` do not seem to fire inside of `XCTestCase`. Pretty perplexed as to why... We didn't have coverage of bandwidth metrics before this change so I don't think this is a blocker, but lmk if either of you feels differently!

Otherwise, we can use  https://github.com/muxinc/mux-stats-sdk-objc/pull/83 to do manual QA looking at the throughput metrics in the dashboard.

dylanjha:
@nidhik I agree with your judgement. Thanks for making this change!

PR #41, opened 2020-02-19
fix request throughput metrics - set type on `MUXSDKRequestBandwidthEvent`

dylanjha:


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -133,7 +133,7 @@ - (void)timeUpdateTimer:(NSTimer *)timer {
                 loadData.requestVideoWidth = nil;
                 loadData.requestVideoHeight = nil;
                 loadData.requestRenditionLists = nil;
-                [self dispatchBandwidthMetric:loadData];
+                [self dispatchBandwidthMetric:loadData withType:MUXSDKPlaybackEventRequestBandwidthEventCompleteType];
```
nidhik: Nice, good idea to add this parameter to the method since it is required for `MuxCore` to properly process the bandwidth metric

PR #36, opened 2020-02-06
[9473] Fix bug in Mux-Stats-AVPlayer that prevents videoChangeForPlayer from working

nidhik:
## Overview

This builds on the work done in https://github.com/muxinc/mux-stats-sdk-avplayer/pull/35

The player binding detects the case in which the video has changed and goes through the `PlayerBindingManager` to make sure the `viewend`, `viewinit`, `data`, and `play` events get sent to `MuxCore` in the correct order. 

https://zube.io/muxinc/mux/c/9473

## Notes
* Easiest to see the diff here: https://github.com/muxinc/mux-stats-sdk-avplayer/pull/36/commits/ac2b781b08a57468d46da39fb8a415e6d893b786
* When testing how changing videos inside of an `AVQueuePlayer` works,  ran across another bug. I filed a ticket here: https://zube.io/muxinc/mux/c/9836 I think it is best fixed separately.

PR #42, opened 2020-02-21
bugfix - programChange and dispatch viewend when player is destroyed

dylanjha:
Two changes here:

1. fix programChange
1. when calling `destroyPlayer`, let's dispatch `viewend` - it looks like an oversight that we were not doing this

----

When calling programChange, we need to call monitorAVPlayerItem on the playerBinding instance so that the binding will call videoChangedForPlayer on the playDispatchDelegate, otherwise the program change will not be properly tracked

QA:

1. Try using `programChangeForPlayer` on current master (v1.2.1) - it does not work properly. It ends up sending the first view twice 😑 
1. Try using `programChangeForPlayer` on this branch - it should work now

I'm especially keen on getting @nidhik 's review of this implementation because this touches some stuff she worked on in #35 and #36 

TODO for @dylanjha:

- [x] write a test for this


apps/DemoApp/DemoApp/ViewController.m
```diff
@@ -95,15 +95,22 @@ - (AVPlayer *)testAVQueuePlayer {
...
                                                        selector:@selector(changeVideo:)
                                                        userInfo:nil
                                                         repeats:NO];
+
+//    // After 20 seconds, we'll change the program
+//    _videoChangeTimer = [NSTimer scheduledTimerWithTimeInterval:20.0
+//                                                         target:self
+//                                                       selector:@selector(changeProgram:)
+//                                                       userInfo:nil
+//                                                        repeats:NO];
```
dylanjha: added this commented code to the demo app so we can test changeProgram

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -231,6 +231,7 @@ + (void)programChangeForPlayer:(nonnull NSString *)name withVideoData:(nullable
     [MUXSDKStats videoChangeForPlayer: name withVideoData:videoData];
     MUXSDKPlayerBinding *player = [_viewControllers valueForKey:name];
     if (player) {
+        [player monitorAVPlayerItem];
```
nidhik: This works and my QA is showing the program change correctly as 2 different views. 
1. `MUXSDKStats:programChangeForPlayer` is called
1. `MUXSDKStats:videoChangeForPlayer` updates the cached video metadata
1. `MUXSDKPlayerBinding:monitorAVPlayerItem` results in a call to `MUXSDKPlayerBindingManager:videoChangedForPlayer` which dispatches `viewinit` and `Data Event` with updated metadata
1. We manually dispatch `play` and `playing` since in this case KVO doesn't fire

My only comment is that perhaps we should add a method `programChangedForPlayer` on `MUXSDKPlayerBinding` and have it call `monitorAVPlayerItem` `dispatchPlay` and `dispatchPlaying` itself - so we don't need to expose `monitorAVPlayerItem` in the public header.

dylanjha: I like that suggestion. I'm going to try to add a test for this too so I'll do that

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -286,6 +286,12 @@ - (void)stopMonitoringAVPlayerItem {
     _playerItem = nil;
 }
 
+- (void) programChangedForPlayer {
```
nidhik: 👍

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -199,10 +199,12 @@ + (void)destroyPlayer:(NSString *)name {
     NSString *binding = [_bindings valueForKey:name];
     if (binding == MuxPlayerSoftwareAVPlayerViewController) {
         MUXSDKAVPlayerViewControllerBinding *player = [_viewControllers valueForKey:name];
+        [player dispatchViewEnd];
```
nidhik: nice catch

MUXSDKStats/MUXSDKStatsTests/MUXSDKStatsTests.m
```diff
@@ -96,6 +137,25 @@ - (void)testVideoChangeForAVPlayerLayer{
...
+    MuxMockAVPlayerViewController *controller = [[MuxMockAVPlayerViewController alloc] init];
+    MUXSDKCustomerPlayerData *customerPlayerData = [[MUXSDKCustomerPlayerData alloc] initWithEnvironmentKey:@"ENV_KEY"];
+    MUXSDKCustomerVideoData *customerVideoData = [[MUXSDKCustomerVideoData alloc] init];
+    [customerVideoData setVideoTitle:@"01234"];
+    NSString *playName = @"Player";
+    MUXSDKPlayerBinding *playerBinding = [MUXSDKStats monitorAVPlayerViewController:controller withPlayerName:playName playerData:customerPlayerData videoData:customerVideoData];
+    XCTAssertNotNil(playerBinding, "expected monitorAVPlayerViewController to return a playerBinding");
+    [customerVideoData setVideoTitle:@"56789"];
+    [MUXSDKStats programChangeForPlayer:playName withVideoData:customerVideoData];
+    NSArray *expectedEventTypes = @[MUXSDKPlaybackEventViewInitEventType,
```
nidhik: Shouldn't we have more events here?
```
// monitor
MUXSDKPlaybackEventViewInitEventType,
MUXSDKDataEventType,
MUXSDKPlaybackEventPlayerReadyEventType,

// program change
MUXSDKPlaybackEventViewEndEventType
MUXSDKPlaybackEventViewInitEventType,
MUXSDKDataEventType,
MUXSDKPlaybackEventPlayerReadyEventType,
MUXSDKPlaybackEventPlayEventType
MUXSDKPlaybackEventPlayingEventType
```

dylanjha: @nidhik good catch! I just updated and updated the asserts with some more friendly error messages and also assert that the number of events matches the expected

nidhik: Thanks 😄

PR #43, opened 2020-02-26
ignore AVPlayer access log notifications if they are for a different item than the one we are monitoring

dylanjha:


dylanjha:
Yeah I realized that. We could initialize a notification object and trigger the method ourselves, right?

nidhik:
@dylanjha That's a good point. I wonder if we initialized an `AVPlayer` from inside the test case, started playback and then manually created the notification with the currently playing `AVPlayerItem`, if the item's `accessLog` would be correctly populated. Seems worth a try

PR #44, opened 2020-02-28
requestCompletedTime is in seconds, not ms, do our with with all seconds values, then multiply by 1000 to get ms

dylanjha:


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -168,9 +168,10 @@ - (void) calculateBandwidthMetricFromAccessLog:(AVPlayerItemAccessLog *) log {
        // but if use data stored in requestResponseStart/requestResponseEnd/requestBytesLoaded to compute, the value are very close.
        MUXSDKBandwidthMetricData *loadData = [[MUXSDKBandwidthMetricData alloc] init];
        loadData.requestType = @"media";
-       loadData.requestStart = [NSNumber numberWithLong: (long)(requestCompletedTime - (event.transferDuration - _lastTransferDuration) * 1000)];
+       double requestStartSecs = requestCompletedTime - (event.transferDuration - _lastTransferDuration);
```
dylanjha: `requestCompletedTime` variable
   * this comes from [`[[NSDate date] timeIntervalSince1970]`](https://developer.apple.com/documentation/foundation/nsdate/1407504-timeintervalsince1970) which is a [`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval)

`event.transferDuration` comes from [`AVPlayerItemAccessLogEvent`](https://developer.apple.com/documentation/avfoundation/avplayeritemaccesslogevent) which is also a [`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval).

`_lastTransferDuration` is also a [`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval)

----

`TimeInterval` is:

  * A TimeInterval value is always specified in seconds; it yields sub-millisecond precision over a range of 10,000 years.

----

When we have these 3 values, they are always **seconds** but they have millisecond precision. So we should do all that math, ultimately end up with a unix timestamp in seconds and multiply it by 1000 to convert the unix timestamp into milliseconds.

---

Ultimately, we end up setting:

`loadData.requestStart` and `loadData. requestResponseEnd` to be unix **millisecond** timestamps.

PR #47, opened 2020-04-02
Dj/av queue player

dylanjha:
Fix for AVQueuePlayer. 

This is how AVQueuePlayer should be used (we never really documented this):

1. Application listens for `AVPlayerItemDidPlayToEndTime` (first item ends) 
1. Application tells Mux `MUXSDKStats.videoChange` - with new videoData (optionally new playerData too, if they want)
1. That should "just work"... but right now it doesn't. See the code below for an example application implementation.

```swift
let playName = "iOS AVPlayer"

override func viewDidLoad() {
    super.viewDidLoad()

    let item1 = AVPlayerItem(url: URL(string: "https://stream.mux.com/jY02nK1sxQKmJiQ7ltXY01w9LZQWdtNetE.m3u8")!)
    let item2 = AVPlayerItem(url: URL(string: "https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8")!)
    NotificationCenter.default.addObserver(
        self,
        selector: #selector(self.playerItemDidReachEnd),
        name: NSNotification.Name.AVPlayerItemDidPlayToEndTime,
        object: item1
    )
    player = AVQueuePlayer(items: [item1, item2])

    let playerData = MUXSDKCustomerPlayerData(environmentKey: "ENV_KEY");
    playerData?.playerName = "AVPlayer"
    let videoData = MUXSDKCustomerVideoData();
    videoData.videoIsLive = false;
    videoData.videoTitle = "Title1"
    MUXSDKStats.monitorAVPlayerViewController(self, withPlayerName: playName, playerData: playerData!, videoData: videoData);
    player!.play()
}

@objc func playerItemDidReachEnd (notification: NSNotification) {
    let videoData = MUXSDKCustomerVideoData();
    videoData.videoTitle = "Title2"
    videoData.videoId = "applekeynote2010-2"
    MUXSDKStats.videoChange(forPlayer: playName, with: videoData)
}
```

-----

The reason this didn't work is because our SDK will:

1. see that the item in the player changes
1. end the first view
1. initialize a second view
1. wait for the player KVO `rate` to change

The problem is... Step 4 never happens in `AVQueuePlayer` world. Note that when calling something like `avPlayer.replaceCurrentItem(with: item2)` the KVO for `rate` *does* fire....

Is this a bug? Is this the intended behavior? (I have no fucking idea)

---

**Solution**

When `videoChangeForPlayer` is called - we internally check if we're dealing with `AVQueuePlayer`. If we are, then we set a flag so that at initialization time we will `dispatchPlay`

---

**QA steps**

- [x] pull down this branch
- [x] run video-demo - check the `testAvPlayer()` method and `testAvQueuePlayer` methods - see that they both work


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -285,6 +285,21 @@ - (void)monitorAVPlayerItem {
...
+        // In a normal videoChange: world - the KVO for "rate" will fire - and
+        // subsequently after that this binding will dispatchPlay. In fact, any time
+        // an AVPlayer gets an item loaded into it the KVO for "rate" changes.
+        //
+        // However, in AVQueuePlayer world - the "rate" doesn't fire when the video is
+        // changed. I don't know why, but I guess that is the intended behavior. For that
+        // reason, if we're handling a videoChange event and we're dealing with AVQueuePlayer
+        // then we have to fire the play event here.
+        //
+        if (_shouldHandleAVQueuePlayerItem) {
```
nidhik, line 299: 👍

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -221,6 +221,7 @@ + (void)videoChangeForPlayer:(nonnull NSString *)name withPlayerData:(nullable M
             if (playerData) {
                 [_customerPlayerDataStore setPlayerData:playerData forPlayerName:name];
             }
+            [player prepareForAvQueuePlayerNextItem];
```
nidhik, line 224: Oh I see, you actually don't check for `AVQueuePlayer` out here, you check inside of `prepareForAvQueuePlayerNextItem`

apps/video-demo/video-demo/VideoPlayerController.swift
```diff
@@ -9,37 +9,61 @@ class VideoPlayerController: AVPlayerViewController {
...
-                                                           queue: DispatchQueue.main) { time in
-                                                            let timeElapsed = Float(CMTimeGetSeconds(time))
-                                                            print ("current time " + timeElapsed.description)
-        }
-        playComplete = NotificationCenter.default.addObserver(self, selector: #selector(self.playerDidFinishPlaying(note:)),
-                                               name: NSNotification.Name.AVPlayerItemDidPlayToEndTime, object: player!.currentItem)
         player!.play()
     }
+    
+    func testAvPlayer () -> AVPlayer {
```
nidhik, line 32: Thanks for updating the demo app

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
dylanjha: yeah this top level MUXSDKStats class doesn't have access to the underlying `player` instance... And that felt more like the responsibility of the binding to handle internally

nidhik: 💯

MUXSDKStats/Podfile.lock
```diff
@@ -13,4 +13,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 424113a5d70474d84755984dcd441bbebc595369
 
-COCOAPODS: 1.9.1
```
dylanjha, line 16: lol okay fine I'll update cocoapods, thanks Podfile

PR #49, opened 2020-05-01
rebuild frameworks without UIKit

dylanjha:
* we don't use it

---

This is causing a reference to UIWebView to be included in the debug symbols for our framework, which is being deprecated

> The App Store will no longer accept new apps using UIWebView as of
April 2020 and app updates using UIWebView as of December 2020.

----

I think this worked, see the following:

In the DemoApp, from master:

```
❯ grep -r "UIWebView" .
Binary file ./Pods/Mux-Stats-AVPlayer/Frameworks/iOS/fat/MUXSDKStats.framework.dSYM/Contents/Resources/DWARF/MUXSDKStats matches
Binary file ./Pods/GoogleAds-IMA-iOS-SDK/GoogleInteractiveMediaAds.framework/GoogleInteractiveMediaAds matches
```

In the DemoApp, from this branch:

```
❯ grep -r "UIWebView" .
Binary file ./Pods/GoogleAds-IMA-iOS-SDK/GoogleInteractiveMediaAds.framework/GoogleInteractiveMediaAds matches
```


MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -9,7 +9,6 @@
 @import AVFoundation;
 @import AVKit;
 @import Foundation;
-@import UIKit;
```
dylanjha, line 12: this is the only thing that changed

PR #56, opened 2020-07-01
Reduces main thread usage by calling `getSourceDimensions` fewer times

nidhik:
## Zube
https://zube.io/muxinc/mux/c/11962

## Overview

Although I would have liked to only get source dimensions on `renditionchange`, unfortunately we may not have the most up to date dimension information at that time. As far as I can tell, `AVFoundation` also doesn't give us a way to know if we have the most up to date dimension information. The dimensions will just eventually be correct at some unknown point in the future 😞 

So we set a boolean flag on `renditionchange` indicating that we know source dimensions have probably changed, then continue to make the expensive `getSourceDimensions` call every 0.1 seconds until we have new dimensions. Once we have source dimensions that are new, we reset the flag and stop making the expensive call until it is once again set to true.

This approach is far from efficient and is actually kind of ugly, but it does improve the performance measurably. It's unfortunate that it adds complexity in exchange for the speed improvement. I'm definitely open to other ideas/approaches.

## Results

With this change, now `MUXSDKCore` takes more time than `checkVideoData` in `MUXSDKStats`. So we have effectively cut down the time we spend getting dimensions enough that a different part of the code is the heaviest. 

Screen Shot 2020-07-01 at 2 06 37 AM

And inside of `checkVideoData` now `[player.currentItem.asset duration]` takes far more time than getting the source dimensions. 

(Checking if the `duration` property is even loaded is second most expensive, so for anyone trying to improve performance in the future, I'd start with `MUXSDKCore` rather than trying to tackle this) 

Screen Shot 2020-07-01 at 3 19 03 AM

## TODO

- [x] Update unit tests now that we send fewer data events
- [x] Remove all debug logging
- [x] QA `changeVideo` 
- [x] QA Ads


MUXSDKStats/Podfile.lock
```diff
@@ -13,4 +13,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 424113a5d70474d84755984dcd441bbebc595369
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.8.4
```
nidhik, line 16: note to self, update cocoapods

apps/DemoApp/DemoApp/ViewController.m
```diff
@@ -137,7 +137,7 @@ - (void)setupAVPlayerViewController:(AVPlayer *)player {
     _avplayerController.player = _avplayer;
 
     // TODO: Add your property key!
-    MUXSDKCustomerPlayerData *playerData = [[MUXSDKCustomerPlayerData alloc] initWithPropertyKey:@"YOUR_ENV_KEY_HERE"];
+    MUXSDKCustomerPlayerData *playerData = [[MUXSDKCustomerPlayerData alloc] initWithPropertyKey:@"is1368012c267a34rk0rhhr4u"];
```
nidhik, line 140: TODO: remove this

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -347,11 +348,13 @@ - (CGRect)getViewBounds {
...
     @try {
+        NSArray *formatDescriptions;
         for (int t = 0; t < _player.currentItem.tracks.count; t++) {
             AVPlayerItemTrack *track = [[[_player currentItem] tracks] objectAtIndex:t];
             if (track) {
-                for (int i = 0; i < track.assetTrack.formatDescriptions.count; i++) {
-                    CMFormatDescriptionRef desc = (__bridge CMFormatDescriptionRef)track.assetTrack.formatDescriptions[i];
+                formatDescriptions = track.assetTrack.formatDescriptions;
+                for (int i = 0; i < formatDescriptions.count; i++) {
+                    CMFormatDescriptionRef desc = (__bridge CMFormatDescriptionRef)formatDescriptions[i];
```
dylanjha, line 357: no functional change here, right?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -400,13 +399,26 @@ - (void)checkVideoData {
     if ([currentPlayerAsset isKindOfClass:AVURLAsset.class]) {
         AVURLAsset *urlAsset = (AVURLAsset *)currentPlayerAsset;
         NSString * urlString = [[urlAsset URL] absoluteString];
-        if (!_videoURL || [_videoURL isEqualToString:urlString]) {
+        if (!_videoURL || ![_videoURL isEqualToString:urlString]) {
```
dylanjha, line 402: unrelated bugfix

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -400,13 +399,26 @@ - (void)checkVideoData {
...
             _videoURL = urlString;
             videoDataUpdated = YES;
         }
     }
-    if (_lastAdvertisedBitrate > 0) {
+    if (![@(_lastDispatchedAdvertisedBitrate) doubleValueIsEqual:@(_lastAdvertisedBitrate)]) {
```
dylanjha, line 407: This is the magic, set `_sourceDimensionsHaveChanged` because we know the advertised bitrate has changed

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -400,13 +399,26 @@ - (void)checkVideoData {
...
+    if (_sourceDimensionsHaveChanged && CGSizeEqualToSize(_videoSize, _lastDispatchedVideoSize)) {
+        CGSize sourceDimensions = [self getSourceDimensions];
+        if (!CGSizeEqualToSize(_videoSize, sourceDimensions)) {
+            _videoSize = sourceDimensions;
+            if (sourceDimensions.width > 0 && sourceDimensions.height > 0) {
+                _lastDispatchedVideoSize = _videoSize;
+                _sourceDimensionsHaveChanged = NO;
+                videoDataUpdated = YES;
+            }
+        }
```
dylanjha, line 421: Now the real magic, if source dimensions have changed then we compare _videoSize to sourceDimensions if they are not equal then we know `sourceDimensions` has changed.

Next, we make sure the `sourceDimensions` value is valid.

And if so then we reset `_sourceDimensionsHaveChanged`

----

The above block will happen repeatedly (every ~0.1s) until we get valid values for `sourceDimensions`, correct? And the reason for that is that we don't know when that value is populated. In AVPlayer world that value is "eventually correct".

That's unfortunate but we have to work with it.

MUXSDKStats/MUXSDKStatsTests/MUXSDKStatsTests.m
```diff
@@ -275,7 +275,7 @@ - (void) testRenditionChangeEvent {
     [[NSNotificationCenter defaultCenter] postNotificationName:RenditionChangeNotification object:renditionInfo];
     
     NSDictionary *renditionInfoWithoutFrameRate = @{
-        RenditionChangeNotificationInfoAdvertisedBitrate: @(258157)
+        RenditionChangeNotificationInfoAdvertisedBitrate: @(558157)
```
dylanjha, line 278: Reason for this bitrate number changing in the test? 😄 nbd, jw

MUXSDKStats/MUXSDKStatsTests/MUXSDKStatsTests.m
```diff
@@ -337,29 +337,27 @@ - (void) testOrientationAndRenditionChangeEventSequence {
...
     [self assertPlayer:playName dispatchedPlaybackEvents:@{
         @(0): [NSNull null],
         @(2): [NSNull null],
         @(3): @{X: @(0.0), Y: @(0.0), Z: @(90.0)},
         @(5): [NSNull null],
         @(7): [NSNull null],
-        @(9): @{X: @(0.0), Y: @(0.0), Z: @(0.0)},
-        @(11): [NSNull null],
+        @(8): @{X: @(0.0), Y: @(0.0), Z: @(0.0)},
+        @(10): [NSNull null],
```
dylanjha, line 360: What changed in this `renditionchange` event?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: Nope, just avoiding calling `track.assetTrack.formatDescriptions` multiple times

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: Though this reduces the amount of work we are doing by a lot :) We used to send a data event every 0.1 sec because of this

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: >The above block will happen repeatedly (every ~0.1s) until we get valid values for sourceDimensions, correct? 

That is correct

MUXSDKStats/MUXSDKStatsTests/MUXSDKStatsTests.m
nidhik: Good eye 😁  It's because of the change I made here: https://github.com/muxinc/mux-stats-sdk-avplayer/pull/56#discussion_r448753922

Previously, we would emit a data event even if the bitrate had not changed if we got two `RenditionChangeNotification`. Now we only do that if the bitrate has changed since the last time we sent it. So I updated the test to reflect that.

Though now that you bring it up, I should also add a unit test covering the case we get two `RenditionChangeNotification` anyway - I'll add this

MUXSDKStats/MUXSDKStatsTests/MUXSDKStatsTests.m
nidhik: We send one fewer DataEvent so the indices changed:
```
    // Assert sequence of playback & data events is correct

    NSArray *expectedEventTypes = @[MUXSDKPlaybackEventViewInitEventType,
                                    MUXSDKDataEventType,
                                    MUXSDKPlaybackEventPlayerReadyEventType,
                                    MUXSDKPlaybackEventOrientationChangeEventType,
                                    MUXSDKDataEventType,
                                    MUXSDKPlaybackEventRenditionChangeEventType,
                                    MUXSDKDataEventType,
                                    MUXSDKPlaybackEventRenditionChangeEventType,
                                    MUXSDKPlaybackEventOrientationChangeEventType,
                                    MUXSDKDataEventType,
                                    MUXSDKPlaybackEventRenditionChangeEventType
```

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -128,7 +129,9 @@ - (void) handleRenditionChange:(NSNotification *) notif {
     NSNumber *advertisedBitrate = renditionChangeInfo[RenditionChangeNotificationInfoAdvertisedBitrate];
     if (advertisedBitrate) {
         _lastAdvertisedBitrate = [advertisedBitrate doubleValue];
-        [self dispatchRenditionChange];
+        if(![@(_lastDispatchedAdvertisedBitrate) doubleValueIsEqual:@(_lastAdvertisedBitrate)]) {
```
nidhik, line 132: I added this since the last review, calling it out. Basically don't dispatch a rendition change event if the bitrate hasn't actually changed.

PR #60, opened 2020-09-09
Viewer connection type

dylanjha:
QA:

- [x] apple TV
- [x] actual hardware on wifi
- [x] actual hardware on cellular

TODO:

- [x] rebuild frameworks after approved


MUXSDKStats/Podfile
```diff
@@ -22,5 +24,6 @@ target 'MUXSDKStatsTv' do
   # use_frameworks!
 
   # Pods for MUXSDKStatsTv
-  pod 'Mux-Stats-Core', '~>2.2.0'
+  #pod 'Mux-Stats-Core', '~>2.2.0'
+  pod 'Mux-Stats-Core', :path => '~/code/@muxinc/mux-stats-sdk-objc/stats-sdk-objc'
```
dylanjha, line 28: when 2.3 of Mux-Core gets released then we can point to that

dylanjha:
iOS - cellular and wifi:

vct_2020-10-01_16-30-02


wifi_2020-10-01_16-30-41

TVOS - wifi:

tvos_2020-10-01_16-34-47

MUXSDKStats/Podfile
dylanjha: done

dylanjha:
A couple things:

- [ ] we must return "ethernet" for a wired apple TV connection
- [ ] do this off the main thread

I was looking at some libs that do this stuff [Reachability](https://github.com/GyazSquare/Reachability#using-the-sample) and noticed this:

>  You can use the API synchronously, but do not issue a synchronous check by hostName on the main thread. If the device cannot reach a DNS server or is on a slow network, a synchronous call to the SCNetworkReachabilityGetFlags function can block for up to 30 seconds trying to resolve the hostName. If this happens on the main thread, the application watchdog will kill the application after 20 seconds of inactivity.

jsanford8 (approved):
lgtm, I think for now we just need to deal with tvos as you do, _or_ just ignore it. I'll leave that decision to Steve Lyons, though, but the code here lgtm

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -117,6 +119,19 @@ - (BOOL) checkIfNotificationIsRelevant:(NSNotification *)notif {
     return notificationItem == _playerItem;
 }
 
+- (void)handleConnectionTypeDetected:(NSNotification *)notif {
+    dispatch_async(dispatch_get_main_queue(), ^{
```
nidhik, line 123: 👍  That's right, we get the notification on a background thread but handle it on the main

nidhik: You may have already thought this through but is it possible we get this notification for a different player item than the current one? I seem to remember this could happen with some other async things and we sometimes use `checkIfNotificationIsRelevant`

MUXSDKStats/MUXSDKStats/MUXSDKConnection.m
```diff
@@ -0,0 +1,57 @@
+#import "MUXSDKConnection.h"
+#import <SystemConfiguration/SystemConfiguration.h>
+
+/*
+    dylanjhaveri
+    Right now this class has 1 significant shortcoming and it is that on
+    a wired ethernet TVOS app it will detect "wifi" as the connection type
+
+    If we can bump the target iOS version for this SDK all the way up to iOS12
```
nidhik, line 9: You may have explained this already and I missed it but was there a reason we couldn't check if iOS12 was available and conditionally use `NWPathMonitor`?

nidhik (approved):
This lgtm with a couple of questions

MUXSDKStats/MUXSDKStats/MUXSDKConnection.m
dylanjha: It's because of the linker when we run `xcodebuild` 😢 

As I understand it, because we're targeting ios 9+, when we run xcodebuild the linker was not building when it saw `nw_path_monitor_create` and things that are only in ios 12+ (even after I had that code surround by obj-c preprocessor macros).

If there is a way to include ios 12+ functionality while still building our SDK for ios 9+, I was unable to do it. Maybe we can with XCFrameworks? Worth trying when we make that change, at least.

PR #62, opened 2020-09-29
Create v1.5.0 release

dylanjha:
- sets min iOS deploy target to 9.0 in podspec and library project build settings for XCode 12 build compatability
- bumps Mux-Stats-Core dependency to ~> 2.3

Note that unlike Mux-Stats-Core, we did not have to do this podspec workaround https://github.com/CocoaPods/CocoaPods/issues/10065 because in this SDK arm_64 was already an excluded architecture from simulators in the build settings .


nidhik (approved):
lgtm

PR #64, opened 2020-10-21
Builds MUXSDKStats as an XCFramework

nidhik:
## Zube
https://zube.io/muxinc/mux/c/11914
https://zube.io/muxinc/mux/c/13151

## Overview

* Builds the MUXSDKStats AVPlayer SDK as a XCFramework
* The product name for `MUXSDKStatsTv` is now also  `MUXSDKStats`
* Supports Catalyst builds. 

## TODO

- [x] Update example demo apps

## QA

- [x] Can use `MUXSDKStats.xcframework` in a Cocoapods/Obj-c project.
- [x] Can use `MUXSDKStats.xcframework` in a Cocoapods/Swift project.
- [x] Can use `MUXSDKStats.xcframework` in a Obj-c regular xcode project.
- [x] Can use `MUXSDKStats.xcframework` in a Swift regular xcode project.

Then

For *each* of the above, we need to test 
- [x] iOS Simulator
- [x] tvOS Simulator
- [x] iPhone hardware
- [x] tvOS hardware

Then ideally for each of these tests we would also make sure we can use the debugger (this will indicate that the dSYM files are being included properly in the projects but also that they are able to be found by the projects integrating our SDK.


.buildkite/build.sh
```diff
@@ -2,7 +2,8 @@
 set -euo pipefail
 
 cd MUXSDKStats
+pod repo update
 pod deintegrate && pod install
 cd ..
-./update-release-frameworks.sh
```
dylanjha, line 7: `./update-release-frameworks` file can be removed

PR #65, opened 2020-10-27
Automated builds for Mux-Stats-AVPlayer

nidhik:
## Zube
https://zube.io/muxinc/mux/c/13151

## Overview

Builds the AVPlayer SDK on a MacStadium mac mini running our Buildkite agent. 

This build: https://buildkite.com/mux/stats-sdk-avplayer/builds/6#796cfc64-e77b-4ed7-9703-d19893e55bd2
Screen Shot 2020-10-27 at 1 27 03 PM

### DIFF
--- a/apps/DemoApp/DemoAppUITests/DemoAppUITests.m
+++ b/apps/DemoApp/DemoAppUITests/DemoAppUITests.m
@@ -29,7 +29,7 @@
 
 - (void)testPlayVideo {
     XCUIApplication *app = [[XCUIApplication alloc] init];
-    [app setLaunchEnvironment:@{@"ENV_KEY": @"is1368012c267a34rk0rhhr4u"}];
+    [app setLaunchEnvironment:@{@"ENV_KEY": @"72j9sdhp80p89p1l659smj69m"}];
     [app launch];
     XCTestExpectation *exp = [[XCTestExpectation alloc] initWithDescription:@"Just wait for 20 seconds."];
     XCTWaiterResult result = [XCTWaiter waitForExpectations:@[exp] timeout:20.0];
--- a/apps/DemoApp/DemoAppUITests/DemoAppUITests.m
+++ b/apps/DemoApp/DemoAppUITests/DemoAppUITests.m
@@ -1,0 +1,41 @@
+//
+//  DemoAppUITests.m
+//  DemoAppUITests
+//
+//  Created by Nidhi Kulkarni on 10/27/20.
+//  Copyright © 2020 Mux, Inc. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+
+@interface DemoAppUITests : XCTestCase
+
+@end
+
+@implementation DemoAppUITests
+
+- (void)setUp {
+    // Put setup code here. This method is called before the invocation of each test method in the class.
+
+    // In UI tests it is usually best to stop immediately when a failure occurs.
+    self.continueAfterFailure = NO;
+    
+    // In UI tests it’s important to set the initial state - such as interface orientation - required for your tests before they run. The setUp method is a good place to do this.
+}
+
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+}
+
+- (void)testPlayVideo {
+    XCUIApplication *app = [[XCUIApplication alloc] init];
+    [app setLaunchEnvironment:@{@"ENV_KEY": @"is1368012c267a34rk0rhhr4u"}];
+    [app launch];
+    XCTestExpectation *exp = [[XCTestExpectation alloc] initWithDescription:@"Just wait for 20 seconds."];
+    XCTWaiterResult result = [XCTWaiter waitForExpectations:@[exp] timeout:20.0];
+    if(result != XCTWaiterResultTimedOut) {
+        XCTFail(@"Interrupted while playing video.");
+    }
+}
+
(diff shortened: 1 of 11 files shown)


skidder (changes requested):
Can you zip the artifacts into a single file?

nidhik:
This shows the artifacts zipped into a single file: https://buildkite.com/mux/stats-sdk-avplayer/builds/22

buildkite.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+cd MUXSDKStats
+pod deintegrate && pod install
+cd ..
+./update-release-frameworks.sh
+zip -r MUXSDKStats.framework.zip Frameworks
+buildkite-agent artifact upload "MUXSDKStats.framework.zip"
```
skidder, line 9: Similar comment to https://github.com/muxinc/mux-stats-sdk-objc/pull/91/files, we should create a `pipeline.yml` config where the build artifacts are specified

skidder (changes requested):

buildkite.sh
nidhik: done here: https://github.com/muxinc/mux-stats-sdk-avplayer/pull/65/commits/508fbf34d1dc4031398cbb14de5e6df12199b609

apps/DemoApp/run-tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+# Delete the old stuff
+rm -Rf Frameworks
+
+buildkite-agent artifact download "MUXSDKStats.framework.zip" . --step "buildkite.sh"
+unzip MUXSDKStats.framework.zip
+cd apps/DemoApp
+pod deintegrate && pod install
```
dylanjha, line 10: Quick sanity check here, making sure I understand the steps for what buildkite is doing here:

1. clones
2. rebuild the frameworks (`buildkite.sh`)
3. open demoapp
4. install local version of the pod that just got built
5. open DemoApp workspace and run `test`

For step 4 to work correctly, it is important that the DemoApp `Podfile` is pointed to the local version of the pod, correct? (instead of cocoapods or a github branch, which we have done in the past).

apps/DemoApp/DemoAppUITests/DemoAppUITests.m
```diff
@@ -0,0 +1,41 @@
...
+    // In UI tests it’s important to set the initial state - such as interface orientation - required for your tests before they run. The setUp method is a good place to do this.
+}
+
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+}
+
+- (void)testPlayVideo {
+    XCUIApplication *app = [[XCUIApplication alloc] init];
+    [app setLaunchEnvironment:@{@"ENV_KEY": @"is1368012c267a34rk0rhhr4u"}];
```
dylanjha, line 32: Which org/env is this? I might suggest that we use a new environment in the same org that we use for [watch-this](https://github.com/muxinc/watch-this)

Instead of hard-coding, it might be nice to load this as an environment variable that gets configured into the build kite process.

dylanjha (approved):

apps/DemoApp/run-tests.sh
nidhik: Yes that is correct - the Podfile must point to the local version of the `Mux-Stats-AVPlayer` pod for this to work as expected
```
pod 'Mux-Stats-AVPlayer', :path => '../..'
```
> instead of cocoapods or a github branch, which we have done in the past
hmm this is a good point 🤔   the tests would not be testing the SDK that we just built if this app points to a Github branch or Cocoapods for the AVPlayer SDK. But when would we do that with the intention of checking in the change?Because when it comes time to merge any of our branches, we'd always want DemoApp to point to the local version of the pod right?

apps/DemoApp/DemoAppUITests/DemoAppUITests.m
nidhik: Ah good call. Yeah this is just my development org but I agree with changing it to the one here: https://github.com/muxinc/watch-this

> Instead of hard-coding, it might be nice to load this as an environment variable that gets configured into the build kite process.
This is a great idea, let me look into how to do this

apps/DemoApp/run-tests.sh
dylanjha: Makes sense, I don't think anything needs to be changed here.

> But when would we do that with the intention of checking in the change? Because when it comes time to merge any of our branches, we'd always want DemoApp to point to the local version of the pod right?

At one point we were pointing the DemoApp to Cocoapods so that if a developer looks around in this app they can easily copy/paste that line into their own app.

I think having the Cocoapods line commented out in the Podfile above this line is nice for that purpose, but overall I agree that it makes sense for the DemoApp to always be pointed to the local source

nidhik: Ahh I see - makes sense

apps/DemoApp/DemoAppUITests/DemoAppUITests.m
nidhik: here is a view being sent to that environment: https://dashboard.mux.com/environments/ioi7qg/views/Y2Kd25ztDBjlLZiWJkS8NYSrajiWwmyAzZo

PR #67, opened 2020-11-07
Add a local HTTP server usable from the XCUITest suite

nidhik:
## Zube
https://zube.io/muxinc/mux/c/13156

## Overview

Uses a local HTTP server to capture beacons sent by our SDK. Relies on the update to Mux Core here: https://github.com/muxinc/mux-stats-sdk-objc/pull/92

Our XCUITest suite can then query the local server in order to verify the beacons received.


apps/TestApp/Podfile
```diff
@@ -0,0 +1,20 @@
+# Uncomment the next line to define a global platform for your project
+platform :ios, '9.0'
+
+target 'TestApp' do
+  use_frameworks!
+  pod 'Mux-Stats-Core', :git => 'https://github.com/muxinc/stats-sdk-objc.git', :branch => 'nidhik/testing-local-server'
```
dylanjha, line 6: this should point to 3.0.1, right?

PR #87, opened 2021-04-02
Update README_INTERNAL.md

nidhik:


dylanjha (approved):
nice 👍

dylanjha:
Need the same in here? https://github.com/muxinc/mux-stats-sdk-objc#readme

nidhik:
> Need the same in here? https://github.com/muxinc/mux-stats-sdk-objc#readme

I don't think so since there is no version number there 🤔  https://github.com/muxinc/stats-sdk-objc/blob/master/Package.swift

PR #88, opened 2021-04-12
Code signing is missing for mac catalyst

nidhik:
## Zube
https://zube.io/muxinc/mux/c/16958

## Overview
The root cause here is the fact that we zip the xcframework artifact in the Buildkite pipeline. Zip does not preserve the symlinks in the Mac Catalyst xcframework. Mac Catalyst is the only target which requires symlinks to work - tvOS and iOS do not, which is why we do not see the problem there. See: https://developer.apple.com/forums/thread/666269?answerId=645764022#645764022

If I build the xcframeworks locally and drag them into a mac catalyst app with "Embed and Sign" selected, the project builds without any code signing errors. Adding the `-y` option to `zip` preserves the symlinks. 

From this branch:
Screen Shot 2021-04-13 at 9 34 42 PM


MUXSDKStats/MUXSDKStats.xcodeproj/project.pbxproj
```diff
@@ -744,7 +743,7 @@
 				BUILD_LIBRARY_FOR_DISTRIBUTION = NO;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = GFDDVAVUVC;
```
nidhik, line 747: This is my personal development team which shouldn't be in the project. I changed tour project to not codesign because the application developer will codesign their project with their own team anyway.
Screen Shot 2021-04-12 at 1 20 54 PM

Package.swift
```diff
@@ -15,7 +15,7 @@ let package = Package(
         .package(
             name: "MuxCore",
             url: "https://github.com/muxinc/stats-sdk-objc.git",
-            from: "3.2.0"
+            .branch("nidhik/codesign")
```
dylanjha, line 18: this was for testing purposes, before merging it should change back to `from: "3.2.1"`

PR #89, opened 2021-04-13
AirPlay rebuffering incorrectly reported as paused

nidhik:
## Zube
https://zube.io/muxinc/mux/c/15463

## Overview
 When AirPlaying, AVFoundation reports the playback rate as 0 when we are actually rebuffering. This does not happen when not using AirPlay. We [detect this as a pause](https://dashboard.mux.com/organizations/i4rhki/environments/av8gn4/views/onyGP8NSmzxBpRsP3NCqkzUDL2I9BwW6mN9a).

This PR observes `isPlaybackBufferEmpty` reported by AVFoundation in order to more accurately detect this scenario.

When the user actually presses pause and then play, there are times when `isPlaybackBufferEmpty` is `true` and `timeControlStatus` is `AVPlayerTimeControlStatusPlaying` - which necessitates the need for the `timeControlStatus` check. 

Here is an example of a AirPlay view with both rebuffering and actual (user initiated) pauses. https://dashboard.mux.com/organizations/i4rhki/environments/av8gn4/views/BV9n7gLuaRr7kliw7VTBKNfYDEi9LjwKpPRn

Note: since the value of `isPlaybackBufferEmpty` can flip to true at some arbitrary point after the playback rate changes to 0, we may under report rebuffering. But it is more accurate than what we have currently.


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -887,6 +900,12 @@ - (void)observeValueForKeyPath:(NSString*) path
         if ([self isPlayerInErrorState]) {
             [self dispatchError];
         }
+    } else if (context == MUXSDKAVPlayerItemPlaybackBufferEmptyObservationContext) {
+        if ([_playerItem isPlaybackBufferEmpty] && _player.timeControlStatus != AVPlayerTimeControlStatusPlaying && [self isPausedWhileAirPlaying]) {
+            // We erroneously detected a pause when in fact we are rebuffering. This *only* happens in AirPlay mode
+            [self dispatchPlay];
+            [self dispatchPlaying];
```
nidhik, line 907: If we detect this scenario, kick ourselves back into the playing state so that we actually call `startBuffering` in the time observer: https://github.com/muxinc/mux-stats-sdk-avplayer/blob/0dd6a8582503fd05f0a9f15a5900d25955da6abf/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m#L73

dylanjha (approved):
Approved. Walked through this logic with @nidhik and we QA'd on DemoApp and on a swift demo app that I created recently. Confirmed this fixes the bug.

PR #90, opened 2021-04-15
Allow disabling built-in videochanges

james-mux:
Allows customers to override any default videochanges we call automatically in the SDK. By default, we want to leave this enabled.

This helps in a case where a customer forces a reload of the same video to the player (e.g, when transition from the background to the foreground. When using AVQueuePlayer, this causes Mux to emit a videochange by default. With this PR, we now have `[MUXSDKStats setAutomaticVideoChange:DEMO_PLAYER_NAME enabled:false];` as an option which can be set after the SDK is initialised, to stop that from happening and keep the same video tracked in a single view.

(XCFramework excluded for now for readability)

### DIFF
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -58,7 +58,6 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
 }
 
 - (BOOL)setAutomaticVideoChange:(BOOL)automaticVideoChange {
-    NSLog(@"MUXSDK-JAMES - YEAH ROY!");
     _automaticVideoChange = automaticVideoChange;
     return _automaticVideoChange;
 }
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -22,6 +22,8 @@ float MUXSDKMaxSecsSeekClockDrift = 0.2f;
 // Number of seconds the playhead has to move from the last known playhead position when
 // restarting play to consider the transition to play a seek. (500ms)
 float MUXSDKMaxSecsSeekPlayheadShift = 0.5f;
+// Default to allowing automaticVideoChange to take place
+bool _automaticVideoChange = true;
 
 // AVPlayer observation contexts.
 static void *MUXSDKAVPlayerRateObservationContext = &MUXSDKAVPlayerRateObservationContext;
@@ -53,6 +55,12 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
 - (BOOL)setAutomaticErrorTracking:(BOOL)automaticErrorTracking {
     _automaticErrorTracking = automaticErrorTracking;
     return _automaticErrorTracking;
+}
+
+- (BOOL)setAutomaticVideoChange:(BOOL)automaticVideoChange {
+    NSLog(@"MUXSDK-JAMES - YEAH ROY!");
+    _automaticVideoChange = automaticVideoChange;
+    return _automaticVideoChange;
 }
 
 - (void)attachAVPlayer:(AVPlayer *)player {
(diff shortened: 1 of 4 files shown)


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -312,31 +319,33 @@ - (void)detachAVPlayer {
...
-        //
-        // However, in AVQueuePlayer world - the "rate" doesn't fire when the video is
-        // changed. I don't know why, but I guess that is the intended behavior. For that
-        // reason, if we're handling a videoChange event and we're dealing with AVQueuePlayer
-        // then we have to fire the play event here.
-        //
-        if (_shouldHandleAVQueuePlayerItem) {
-            _shouldHandleAVQueuePlayerItem = false;
-            [self dispatchPlay];
+    if (_automaticVideoChange) {
```
jsanford8, line 322: One thing that can help with too much nesting, especially for `void` return signatures, is to bail out directly, rather than nest further. Where possible, I tend to try to _not_ nesting, so do something like the following instead:

```objc
if (!_automaticVideoChange) {
  return;
}

if (_playerItem) {
  ...
```

MUXSDKStats/MUXSDKStats/MUXSDKStats.h
```diff
@@ -178,6 +178,17 @@ FOUNDATION_EXPORT
...
 
  */
++ (void)setAutomaticVideoChange:(nonnull NSString *)name enabled:(nullable Boolean *)enabled;
+
+/*!
+ @method      setAutomaticVideoChange:forPlayer:enabled
+ @abstract    Allows default videochange functionality to be disabled.
+ @param       name The name of the player to update
+ @param       enabled Whether to enabled or disable the automatic videochange functionality
+ @discussion  Use this method to disable built in videochange calls when using AVQueuePlayer.. The player name provided must been passed as the name in a monitorPlayer:withPlayerName:andConfig: call. The config provided should match the specifications in the Mux docs at https://docs.mux.com and should set the enabled value to false. The default setting is true. If the name of the player provided was not previously initialized, an exception will be raised.
```
jsanford8, line 188: you have an extra `.` in here on the first line.

jsanford8: Also, given how this is written, I don't think it raises an exception if the player wasn't initialized, I think it just straight up no-ops. Right?

james-mux: 👍  i'll clean the extra `.`. On the exception, this was a question I had myself - I guess a no-op is what we'd prefer?

jsanford8: I'd just copy what the error tracking flag does (and if the docs are wrong on the other one, update the docs there too)

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -22,6 +22,8 @@
 // Number of seconds the playhead has to move from the last known playhead position when
 // restarting play to consider the transition to play a seek. (500ms)
 float MUXSDKMaxSecsSeekPlayheadShift = 0.5f;
+// Default to allowing automaticVideoChange to take place
+bool _automaticVideoChange = true;
```
nidhik, line 26: I think you want this to be a private instance variable on the player binding? https://github.com/muxinc/mux-stats-sdk-avplayer/blob/dc7a70d5569184223d5c45f76091addd641e2421/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h#L66

MUXSDKStats/MUXSDKStats/MUXSDKStats.h
nidhik: I think having a new public method like `setAutomaticVideoChange:name:enabled` is fine but we should also have a discussion about how we want various options/flags that modify our default behaviors to be represented in this SDK. 

We initially added the automatic error tracking flag by adding additional monitor methods:
```
+ (MUXSDKPlayerBinding *_Nullable)monitorAVPlayerViewController:(nonnull AVPlayerViewController *)player withPlayerName:(nonnull NSString *)name playerData:(nonnull MUXSDKCustomerPlayerData *)playerData videoData:(nullable MUXSDKCustomerVideoData *)videoData automaticErrorTracking:(BOOL)automaticErrorTracking;
```

Then later, when we added `viewData`, we needed to add additional public methods that also had this error tracking flag:
```
+ (MUXSDKPlayerBinding *_Nullable)monitorAVPlayerViewController:(nonnull AVPlayerViewController *)player withPlayerName:(nonnull NSString *)name playerData:(nonnull MUXSDKCustomerPlayerData *)playerData videoData:(nullable MUXSDKCustomerVideoData *)videoData viewData: (nullable MUXSDKCustomerViewData *) viewData automaticErrorTracking:(BOOL)automaticErrorTracking;
```

This approach will start to get unscalable as we have more options because we will end up with a large number of overloaded ways to call `monitor`.

Our other option is to do what you've done here which is to add a special setter. But this approach also can also get difficult to for end users of this SDK as more options are added because we will have a setter for each one and internally we will also have to remember to implement consistent behavior if the player isn't initialized etc. 

The most idiomatic way to represent options in Obj-C/Swift APIs is to use [NS_OPTIONS macros in Obj-C](https://developer.apple.com/library/archive/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html#//apple_ref/doc/uid/TP40014150-CH1-SW6) which get translated to the [OptionSet in Swift](https://nshipster.com/optionset/). 

Apple uses this approach heavily so it will be familiar to iOS developers.
```
typedef NS_OPTIONS(NSUInteger, MUXSDKOptions) {
    MUXSDKOptionsNone                    = 0,
    MUXSDKOptionsAutomaticErrorTracking  = 1 << 0,
    MUXSDKOptionsAutomaticVideoChange    = 1 << 1,
};
```
Then you can define methods like this:
```
+ (MUXSDKPlayerBinding *_Nullable)monitorAVPlayerViewController:(nonnull AVPlayerViewController *)player withPlayerName:(nonnull NSString *)name playerData:(nonnull MUXSDKCustomerPlayerData *)playerData videoData:(nullable MUXSDKCustomerVideoData *)videoData options:(MUXSDKOptions) options;
```

Check options like this:
```
(options & MUXSDKOptionsAutomaticErrorTracking) == MUXSDKOptionsAutomaticErrorTracking)
```

Set options like this:
```
options = MUXSDKOptionsAutomaticErrorTracking | MUXSDKOptionsAutomaticVideoChange
```

I believe if we did this and moved error tracking flag to this approach it would be a major version bump so I'm not suggesting we do that right now, just wanted to talk about it.

jsanford8: Yes, I really like this options approach, honestly. This is going to get out of hand, quickly. Part of why we added automaticErrorTracking to the constructor, though, was namely because we needed to be able to disable that for errors that happened _during setup_. I think I'd suggest going with the current proposal, but also follow up with `MUXSDKOptions` as we start to join these together.

nidhik: That sounds good to me!

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
james-mux: This is a good shout - i've updated to make it not nested

MUXSDKStats/MUXSDKStats/MUXSDKStats.h
james-mux: I also love this idea long term! For now, i've tidied up the description but we should revisit this for a major bump at some point. I'll write a ticket for the backlog!

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
james-mux: I tried doing this, but couldn't seem to get it to default to being `true` - maybe i'm missing something obvious?

nidhik: I think once you make it an instance variable, you have to set it to true in the initializer: https://github.com/muxinc/mux-stats-sdk-avplayer/blob/1fe8a2808dc8d502483253fcbbeaa2e6bf3b309f/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m#L43

nidhik (approved):

XCFramework/MUXSDKStats.xcframework/ios-arm64_i386_x86_64-simulator/MUXSDKStats.framework/Headers/MUXSDKStats.h
```diff
@@ -178,6 +178,17 @@ FOUNDATION_EXPORT
...
 
  */
++ (void)setAutomaticVideoChange:(nonnull NSString *)name enabled:(Boolean)enabled;
+
+/*!
+ @method      setAutomaticVideoChange:forPlayer:enabled
+ @abstract    Allows default videochange functionality to be disabled.
+ @param       name The name of the player to update
+ @param       enabled Whether to enabled or disable the automatic videochange functionality
+ @discussion  Use this method to disable built in videochange calls when using AVQueuePlayer.. The player name provided must been passed as the name in a monitorPlayer:withPlayerName:andConfig: call. The config provided should match the specifications in the Mux docs at https://docs.mux.com and should set the enabled value to false. The default setting is true.
```
jsanford8, line 188: there's an extra `.` in here... not the biggest deal

jsanford8 (approved):
lgtm though there's an extra `.` hidden in here

PR #92, opened 2021-04-19
Track seeking correctly on tvOS

nidhik:
## Zube
https://zube.io/muxinc/mux/c/16363

## Overview
Core PR required for this to fully work here: https://github.com/muxinc/mux-stats-sdk-objc/pull/107

## QA
Fast forwarding by long press: https://dashboard.mux.com/organizations/i4rhki/environments/av8gn4/views/WZn02k5uD5rZ6FywgHY27hkmEHkz3l00wd7

Pausing then seeking forward in time by swiping: https://dashboard.mux.com/organizations/i4rhki/environments/av8gn4/views/Am60robIN8N2Veh4nRIBD3CV6lsnebgKjyDy

Seek forward, seek backward:
https://dashboard.mux.com/organizations/i4rhki/environments/av8gn4/views/Lry0224u1ajyVlUGaNu937sMyYTep15wEnz3

Pause & no seek
https://dashboard.mux.com/organizations/i4rhki/environments/av8gn4/views/yb2PEGgcxkxyggS4NsVE4iDMBHzAnj6GmpV


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -824,7 +834,11 @@ - (void)computeDrift {
         (_state == MUXSDKPlayerStatePaused || _state == MUXSDKPlayerStatePlay)) {
         _seeking = YES;
         MUXSDKInternalSeekingEvent *event = [[MUXSDKInternalSeekingEvent alloc] init];
-        [event setPlayerData:[self getPlayerData]];
+        MUXSDKPlayerData *playerData = [self getPlayerData];
+# if TVOS
+        [self setPlayerPlayheadTime:_lastPlayheadTimeMsOnPause onPlayerData:playerData];
+#endif
```
dylanjha, line 840: Is a macro the best way to handle this? My gut instinct here is to wrap this in a conditional that is part of the obj-c runtime code.

Not a huge difference, but just something that sticks out to me

nidhik: That's a fair point - I'll make this a runtime check - it doesn't really matter if this code is included in the SDK that is compiled for iOS instead of tvOS

nidhik: And one more point - in order to do this I'll have to import `UIKit`. In MuxCore, we were already referencing `UIKit` so I did a runtime check there but we were not in the AVPlayer SDK. Anyway, since its already a dependency of MuxCore, I don't think it's a big deal to add this dependency - just calling that out

nidhik: Actually - I'm having second thoughts that we should have a dependency on `UIKit` for either of our SDKs at all. The only place I see where it is used in any other way than to distinguish between tvOS/iOS is when building info for Sentry: https://github.com/muxinc/mux-stats-sdk-objc/blob/a84ea72bbb71be0a1399a16a7d882b48d4ff0f43/MuxCore/MuxCore/Core/MUXSDKCorePlayer.m#L95

This uses the current device's user interface idiom to set `os_family` - which isn't actually accurate. For example, the user interface idiom for a Mac Catalyst app can be either `ipad` OR `mac`. https://developer.apple.com/documentation/uikit/mac_catalyst/choosing_a_user_interface_idiom_for_your_mac_app

Everywhere else we use UIKit, we are just using it to distinguish between the tvOS and iOS platforms which can be done at compile time. And though UIKit is likely used by all of our customers right now - that's not actually a requirement for apps. You could make an entire application without UIKit, especially now that SwiftUI is available. I kind of want to remove the UIKit dependency altogether but curious to get your thoughts @dylanjha @jsanford8

dylanjha (approved):

PR #98, opened 2021-05-06
Build static frameworks

nidhik:
## Zube
https://zube.io/muxinc/mux/c/17482

## Overview 

Adds a new Buildkite step that will output statically linked frameworks that do not use Clang modules. This also works for tvOS by loading the correct header tvOS specific header if we have disabled Clang modules.

Screen Shot 2021-05-11 at 6 10 34 PM

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -14,6 +14,6 @@ steps:
   - command: ".buildkite/build-static.sh"
     label: ":hammer: Build Static"
     artifact_paths:
-        - "MUXSDKStats.xcframework.zip"
+        - "MUXSDKStats-static.xcframework.zip"
     agents:
       queue: "iOS"


.buildkite/build-static.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+cd MUXSDKStats
+pod repo update
+pod deintegrate && pod install
+cd ..
+./update-release-xcframeworks-static.sh
+zip -ry MUXSDKStats.xcframework.zip XCFramework
```
skidder, line 9: Let's add `static` to the resulting ZIP filename:
```suggestion
zip -ry MUXSDKStats-static.xcframework.zip XCFramework
```

.buildkite/pipeline.yml
```diff
@@ -10,3 +10,10 @@ steps:
     label: "Tests"
     agents:
       queue: "iOS"
+  - wait
+  - command: ".buildkite/build-static.sh"
+    label: ":hammer: Build Static"
+    artifact_paths:
+        - "MUXSDKStats.xcframework.zip"
```
skidder, line 17: We should name the artifacts differently to avoid confusion.
```suggestion
        - "MUXSDKStats-static.xcframework.zip"
```

skidder (approved):
LGTM with some minor feedback on the ZIP file naming. So excited to have this in our automated build pipeline!

PR #104, opened 2021-08-05
Replace identifierForVendor with alternative UUID

bmulligan-kg:
Mux Viewer ID should use a random uuid, not IDFV on iOS
## Zube
https://zube.io/muxinc/mux-ios-sdks/c/3

### DIFF
--- a/MUXSDKStats/MUXSDKStats/MUXSDKStats.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKStats.m
@@ -20,6 +20,7 @@
 // Software constants.
 NSString *const MuxPlayerSoftwareAVPlayerViewController = @"AVPlayerViewController";
 NSString *const MuxPlayerSoftwareAVPlayerLayer = @"AVPlayerLayer";
+NSString *const MuxDeviceIDUserDefaultsKey = @"MUX_DEVICE_ID";
 
 
 @implementation MUXSDKStats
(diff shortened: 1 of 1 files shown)


MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -315,6 +316,18 @@ + (void)updateAVPlayerLayer:(AVPlayerLayer *)player withPlayerName:(NSString *)n
...
 }
 
+#pragma mark UUID
+
++ (NSString *)getUUIDString {
+    NSString *uuid = [[NSUserDefaults standardUserDefaults] stringForKey:MuxDeviceIDUserDefaultsKey];
+    if (uuid == nil) {
+        uuid = [[NSUUID UUID] UUIDString];
+        [[NSUserDefaults standardUserDefaults] setObject:uuid forKey:MuxDeviceIDUserDefaultsKey];
+        [[NSUserDefaults standardUserDefaults] synchronize];
```
nidhik, line 326: I had to look this up too because I remember when this method was necessary. However, looks we no longer need to call synchronize: https://developer.apple.com/documentation/foundation/nsuserdefaults/1414005-synchronize?language=objc
> Waits for any pending asynchronous updates to the defaults database and returns; this method is unnecessary and shouldn't be used.

But when I went digging this was actually[ introduced as of iOS 12](https://developer.apple.com/documentation/macos-release-notes/foundation-release-notes?preferredLanguage=occ) and we technically target 9...so all this to say I think we need to keep it but that it will be deprecated at some point

nidhik (approved):

PR #105, opened 2021-08-06
Fix race condition when checking connection type

lmadrigalkonradgroup:
### Root Issue
There is a race condition in MUXSDKStats class, specifically when the monitorAVPlayerViewController method is called. Currently the initSDK method is the one that calls the `detectConnectionType` method from MUXSDKConnection (posts a notification when connection type is identified), this init is called at the very beginning of the method, after that the `attachAVPlayer` method of the new binding is called (from here we start listening for connection type notifications) and at this point the notification of the connection type might have already been broadcasted, which is the case when the connection type isn’t reflected in the dashboard.

### Solution
`detectConnectionType` call has been moved out of MUXSDKStats and now is called when `attachAVPlayer` from `MUXSDKPlayerBinding` is called, right after the binding starts listening for the connection type notification.

### Views
Before fix: https://dashboard.mux.com/organizations/jfd5eh/environments/4eao9c/views/bNZM4qquKz9VjRhNWdtYAkcw3ZIDmZ4oZjbA

With fix applied:
https://dashboard.mux.com/organizations/jfd5eh/environments/4eao9c/views/eNjZVgPcrm8PzGtJDZUzoVCdbZsNG9arlDp2

## Zube
https://zube.io/muxinc/mux-ios-sdks/c/2

### DIFF
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -109,6 +109,14 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleAVPlayerError:) name:AVPlayerItemNewErrorLogEntryNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleConnectionTypeDetected:) name:@"com.mux.connection-type-detected" object:nil];
     
+    //
+    // dylanjhaveri
+    // See MUXSDKConnection.m for the tvos shortcoming
+    //
+    if ([[UIDevice currentDevice] userInterfaceIdiom] != UIUserInterfaceIdiomTV) {
+        [MUXSDKConnection detectConnectionType];
+    }
+    
     _lastTransferEventCount = 0;
     _lastTransferDuration= 0;
     _lastTransferredBytes = 0;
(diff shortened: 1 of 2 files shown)


bmulligan-kg:
I think `master` is the incorrect target branch based on our conversation earlier today

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -109,6 +109,14 @@ - (void)attachAVPlayer:(AVPlayer *)player {
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleAVPlayerError:) name:AVPlayerItemNewErrorLogEntryNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleConnectionTypeDetected:) name:@"com.mux.connection-type-detected" object:nil];
     
+    //
+    // dylanjhaveri
+    // See MUXSDKConnection.m for the tvos shortcoming
+    //
+    if ([[UIDevice currentDevice] userInterfaceIdiom] != UIUserInterfaceIdiomTV) {
```
bmulligan-kg, line 116: i think you mentioned that this should be changed to an OS check instead?

lmadrigalkonradgroup: Not exactly, what I said is that the code block in the `initSDK` method is defining 2 variables deviceCategory and osFamily, and the deviceCategory was being used instead of osFamily to do the check, so osFamily should be the one used, but it can really be simplified to this check, basically detect the connection if the device is not an Apple TV.

bmulligan-kg (approved):

nidhik (approved):

PR #108, opened 2021-08-11
Fix crash in video-demo when testing without ads

bmulligan-kg:


apps/video-demo/video-demo/VideoPlayerController.swift
```diff
@@ -94,11 +94,13 @@ class VideoPlayerController: AVPlayerViewController, IMAAdsLoaderDelegate, IMAAd
...
 
     func requestAds() {
+        guard let adsLoader = adsLoader else { return }
+
         let adDisplayContainer = IMAAdDisplayContainer.init(adContainer: self.view, viewController: self)
         let request = IMAAdsRequest.init(
             adTagUrl: adUrl,
             adDisplayContainer: adDisplayContainer,
-            contentPlayhead: (contentPlayhead as! IMAContentPlayhead & NSObjectProtocol),
+            contentPlayhead: (contentPlayhead as? IMAContentPlayhead & NSObjectProtocol),
```
bmulligan-kg, line 103: @nidhik i noticed that contentPlayhead is only initialized for the ad test, but I couldn't determine why (it doesn't _seem_ like it is ad specific). Either way, it's an optional param to this initializer so there's no reason for the force cast

lmadrigalkonradgroup (approved):

nidhik (approved):

PR #116, opened 2021-09-03
[ch712] test AVQueuePlayer

StephanieZuniga:
[ch712] Issue: mux-stats-sdk-avplayer does not create new view when video changes in AVQueuePlayer.

Solution:
- Confirmed that AVQueuePlayer is working as expected.
- Fix DemoApp implementation for testAVQueuePlayer to call MUXSDKStats videoChangeForPlayer: withVideoData when getting the AVPlayerItemDidPlayToEndTimeNotification.
- Add UITest case for testAVQueuePlayer.

### DIFF
--- a/apps/DemoApp/DemoAppUITests/DemoAppUITests.m
+++ b/apps/DemoApp/DemoAppUITests/DemoAppUITests.m
@@ -50,7 +50,7 @@ NSString *const kAdTagURLStringPostRoll = @"https://pubads.g.doubleclick.net/gam
         XCTFail(@"Interrupted while playing video.");
     }
         
-    XCUIElement *element = [[[app.windows childrenMatchingType:XCUIElementTypeOther].element childrenMatchingType:XCUIElementTypeOther].element childrenMatchingType:XCUIElementTypeOther].element;
+    XCUIElement *element = app.otherElements[@"AVPlayerView"];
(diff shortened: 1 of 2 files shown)


nidhik (approved):

apps/DemoApp/DemoAppUITests/DemoAppUITests.m
```diff
@@ -67,6 +67,36 @@ - (void)testIMASDK {
     }
 }
 
+- (void)testAVQueuePlayer {
+    XCUIApplication *app = [[XCUIApplication alloc] init];
+    [app setLaunchEnvironment:@{@"ENV_KEY": @"tr4q3qahs0gflm8b1c75h49ln", @"TEST_SCENARIO": @"AV_QUEUE"}];
```
bmulligan-kg, line 72: your ENV key shouldn't be hardcoded in the tests

StephanieZuniga: This is not my environment key is the one all tests are using , it's this dashboard https://dashboard.mux.com/organizations/ffhei0/environments/k0vjj8/metrics but we or at least I didn't have access. @nidhik was going to look into this to see if we could get permissions. But yeah all tests are targeting that env.

PR #122, opened 2021-11-10
Bmulligan/latency

bmulligan-kg:
Add latency metric reporting


apps/DemoApp/DemoApp/ViewController.m
```diff
@@ -234,7 +245,7 @@ - (void)setupAVPlayerViewController:(AVPlayer *)player {
     // TODO: Add your property key!
     NSString *envKey = [NSProcessInfo.processInfo.environment objectForKey:@"ENV_KEY"];
     if(envKey == nil) {
-        envKey = @"YOUR_ENV_KEY_HERE";
+        envKey = @"clkgm13uf3b3j7ggat0qui8r3";
```
StephanieZuniga, line 248: Is this your key?

apps/DemoApp/DemoAppUITests/DemoAppUITests.m
```diff
@@ -17,7 +17,7 @@ @interface DemoAppUITests : XCTestCase
 @implementation DemoAppUITests
 
 // Set this key to your environment key to have the tests generate data on your dashboard
-static NSString *envKey = @"tr4q3qahs0gflm8b1c75h49ln";
+static NSString *envKey = @"clkgm13uf3b3j7ggat0qui8r3";
```
StephanieZuniga, line 20: I think you left your key?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -231,6 +233,12 @@ - (void) calculateBandwidthMetricFromAccessLog:(AVPlayerItemAccessLog *) log {
     }
 }
 
+- (void) updateViewingLivestream:(AVPlayerItemAccessLog *) log {
+    AVPlayerItemAccessLogEvent *lastEvent = log.events.lastObject;
+    NSString *playbackType = lastEvent.playbackType;
+    _playbackIsLivestream = [playbackType isEqualToString:@"LIVE"];
```
nidhik, line 239: oh nice, so this value is `VOD` if on demand?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -581,10 +589,37 @@ - (MUXSDKPlayerData *)getPlayerData {
                                                                     encoding:NSUTF8StringEncoding]];
         }
     }
+
+    // Only report sampling data if the playback is live
```
nidhik, line 593: let's update this comment to say `Only report program time metrics if playback is live` (since we no longer have a sampling event)

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -581,10 +589,37 @@ - (MUXSDKPlayerData *)getPlayerData {
...
+
+        if ([_player.currentItem.seekableTimeRanges count] > 0) {
+            // seekableTimeRanges is ordered, so we only need to look at the last one
+            // Note about seekableTimeRanges: the meaning of the values in seekableTimeRanges appears to change
+            // across OS versions and device types. Sometimes the duration appears to take holdbacks from the HLS
+            // manifest into consideration, and other times not. Use this value with caution.
+            CMTimeRange seekableRange = [_player.currentItem.seekableTimeRanges.lastObject CMTimeRangeValue];
+            CGFloat start = CMTimeGetSeconds(seekableRange.start);
+            CGFloat duration = CMTimeGetSeconds(seekableRange.duration);
+            CGFloat timeOfVideo = CMTimeGetSeconds([[_player currentItem] currentTime]);
```
nidhik, line 608: What do you think about the name `currentTime`?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -581,10 +589,37 @@ - (MUXSDKPlayerData *)getPlayerData {
...
+            CGFloat livePosition = start + duration;
+            NSTimeInterval viewStartDate = currentDate - timeOfVideo;
+            NSTimeInterval newestProgramDate = viewStartDate + livePosition;
```
nidhik, line 611: I think we should call this `liveEdgeProgramTime` rather than `newestProgramDate` to be consistent with the name of the metric. These are relatively minor nits but since the definitions of these values is tricky, feels worth being extra pedantic about the naming :)

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -581,10 +589,37 @@ - (MUXSDKPlayerData *)getPlayerData {
...
+
+            playerData.playerLiveEdgeProgramTime = [NSNumber numberWithLongLong:(long long)(newestProgramDate * 1000)];
```
nidhik, line 613: I think I'm slightly confused about this calculation, perhaps because we are adding CGFloats to NSDates without using the `addingTimeInterval` API? It might return the same result but having trouble parsing this. This is how I was calculating it:

```
let start = timeRange.start.seconds
let totalDuration = timeRange.duration.seconds
let currentTime = curPlayer.currentTime().seconds
let liveEdge = start + totalDuration
let secondsBehindLive = currentTime - liveEdge
print("  [LATENCY METRIC] Live Edge PDT: ", curDate.addingTimeInterval(-1*secondsBehindLive))
```

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -581,10 +589,37 @@ - (MUXSDKPlayerData *)getPlayerData {
...
+        }
+    }
+
     // TODO: Airplay - don't set the view if we don't actually know what is going on.
+
```
nidhik, line 618: extra ws

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -581,10 +589,37 @@ - (MUXSDKPlayerData *)getPlayerData {
...
     if (_player.externalPlaybackActive) {
         [playerData setPlayerRemotePlayed:[NSNumber numberWithBool:YES]];
     }
+
```
nidhik, line 622: ws

apps/DemoApp/DemoAppUITests/DemoAppUITests.m
```diff
@@ -43,6 +43,17 @@ - (void)testPlayVideo {
     }
 }
 
+- (void)testPlayLivestreamVideo {
+    XCUIApplication *app = [[XCUIApplication alloc] init];
+    [app setLaunchEnvironment:@{@"ENV_KEY": envKey, @"TEST_SCENARIO": @"LIVESTREAM"}];
```
nidhik, line 48: thanks for adding this! Perhaps let's call it `LL-HLS` to distinguish from "regular live"?

apps/video-demo/video-demo/VideoPlayerController.swift
```diff
@@ -28,7 +28,7 @@ class VideoPlayerController: AVPlayerViewController, IMAAdsLoaderDelegate, IMAAd
 //        player = self.testAvQueuePlayer()
 //        player = self.testImaSDK()
 
-        let playerData = MUXSDKCustomerPlayerData(environmentKey: "ENV_KEY")
+        let playerData = MUXSDKCustomerPlayerData(environmentKey: "clkgm13uf3b3j7ggat0qui8r3")
```
nidhik, line 31: key is still in here

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: Actually, a more accurate comment would be `Only report program time metrics if this is a live stream`

apps/DemoApp/DemoApp/ViewController.m
bmulligan-kg: oops

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
bmulligan-kg: yes, or the key could be absent

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
bmulligan-kg: yeah, i think in general there's a lot of confusion with the way we've chosen to name things. I think we use "time" to refer to dates or a number of seconds into a video. I was trying to avoid that confusion here by using "date" in variable names that represent dates, and "time" to represent numbers of seconds. I can change it to match the metric name but i'd rather change the metric name hahah

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
bmulligan-kg: the calculation here is:
- find the timestamp that represents the beginning of the video ([currentItem currentDate] - timeOfVideo). Our understanding is that currentDate is calculated using the X-PROGRAM-DATE from the manifest
- Add the number of seconds that seekableTimeRanges tells us are available for seeking (livePosition) to that timestamp to calculate the timestamp for the video "live edge"
This calculation is only using values from the video itself

The calculation you have is assuming a relationship between the device date (assuming that is what curDate means) and the video dates which I'm not sure will be true

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
bmulligan-kg: maybe i'll rename these "timestamp" instead of "date", i think that's a happy medium

apps/DemoApp/DemoAppUITests/DemoAppUITests.m
bmulligan-kg: i'll add another test that is LL

nidhik:
Summarizing what we talked about via Zoom - the calculations we both came up with for the live edge are the same and only differ as to when we convert to program time

nidhik (approved):

PR #125, opened 2021-11-30
missing seek events

StephanieZuniga:
Add support to track seek events called programatically.


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h
```diff
@@ -47,7 +47,7 @@ typedef NS_ENUM(NSUInteger, MUXSDKViewOrientation) {
     AVPlayer *_player;
     AVPlayerItem *_playerItem;
     id _timeObserver;
-    MUXSDKPlayerState _state;
+    volatile MUXSDKPlayerState _state;
```
bmulligan-kg, line 50: never a good sign when we need to use this lol

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -880,20 +880,35 @@ - (void)computeDrift {
...
+                [self setPlayerPlayheadTime:_lastPlayheadTimeMsOnPause onPlayerData:playerData];
+            }
+            [event setPlayerData:playerData];
+            [MUXSDKCore dispatchEvent:event forPlayer:_name];
+        } else if (_state == MUXSDKPlayerStatePlaying) {
+            // Programatically called seek
+            _seeking = YES;
+            MUXSDKInternalSeekingEvent *seekingEvent = [[MUXSDKInternalSeekingEvent alloc] init];
+            MUXSDKPlayerData *playerData = [self getPlayerData];
+            [self setPlayerPlayheadTime:_lastPlayheadTimeMs onPlayerData:playerData];
```
bmulligan-kg, line 903: is there a reason that this doesn't match lines 893-895?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -880,20 +880,35 @@ - (void)computeDrift {
...
+            _seeking = YES;
+            MUXSDKInternalSeekingEvent *event = [[MUXSDKInternalSeekingEvent alloc] init];
+            MUXSDKPlayerData *playerData = [self getPlayerData];
+            if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomTV) {
+                [self setPlayerPlayheadTime:_lastPlayheadTimeMsOnPause onPlayerData:playerData];
+            }
+            [event setPlayerData:playerData];
+            [MUXSDKCore dispatchEvent:event forPlayer:_name];
+        } else if (_state == MUXSDKPlayerStatePlaying) {
+            // Programatically called seek
```
bmulligan-kg, line 899: couldn't the system programatically call seek if the player was paused as well? this comment implies that the state is playing for any programatic seek

StephanieZuniga: Yes it could be called on pause my intention with this comment was more oriented to the fact that this block of code would only enter if seek was called programatically.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
StephanieZuniga: Yes those lines of code were added because of the way the controller works on apple tv, when you're interacting with the apple tv to seek it pauses and then it was taking the incorrect playhead time. But the only way to get in this block is by calling seek programmatically so there is no special behavior on apple tv cause the seek just happens immediately that's why we send the seeking time as the last playhead time registered and then the seeked time is the current time.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h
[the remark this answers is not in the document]
StephanieZuniga: It's really more of a precaution cause when I was debugging I noticed this variable is constantly set like the breakpoint for setting its value kept pausing in a loop. And I noticed some weird cases where the check for the state was going into the incorrect block of code cause when it checked the conditional state had different value than when performing the logic inside the block. Once I added volatile I didn't see this behavior anymore because it avoids the compilers optimizations that could cause us to get the previous value instead.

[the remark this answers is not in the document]
nidhik: 😬  got it

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -880,20 +880,35 @@ - (void)computeDrift {
     float playheadTimeElapsed = ([self getCurrentPlayheadTimeMs] - _lastPlayheadTimeMs) / 1000;
     float wallTimeElapsed = CFAbsoluteTimeGetCurrent() - _lastPlayheadTimeUpdated;
     float drift = playheadTimeElapsed - wallTimeElapsed;
+
```
nidhik, line 883: extra ws

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: Do you actually need this line? I'm not sure the playhead time has to be explicitly be set here. It does in the case of apple tv because we are attempting to override the playhead with what it was on pause

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: Thanks for explaining this - could we expand this comment a bit to say `when programmatically calling seek, it's possible the player never goes through the paused state` so that it is clear you could also go through the top branch if seeking programmatically

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
StephanieZuniga: I left this ws intentionally because the next few lines have a lot of documentation about the logic and I actually thought the separation here makes it more readable, but I can always remove the ws if you don't agree.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
StephanieZuniga: yes we need to set it because the seek happens immediately which means the seeking playhead would take the seeked playhead time from when the seeking ended instead of the start seeking one that's why we register the lastPlayheadTime and send that for the Seeking instead of leaving the current playhead time we get from playerData which is actually the seeked time.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: ah fair enough - we can leave it in

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
StephanieZuniga: Yeah I can expand

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: ooo I see

nidhik (approved):

PR #126, opened 2021-12-15
Fix picture in picture ads compatibility

StephanieZuniga:
Avoid sending playback events when ad is playing.
Ignore avPlayer item change when ad is playing to fix issue with splitting views for each ad.


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h
```diff
@@ -47,7 +47,7 @@ typedef NS_ENUM(NSUInteger, MUXSDKViewOrientation) {
     AVPlayer *_player;
     AVPlayerItem *_playerItem;
     id _timeObserver;
-    MUXSDKPlayerState _state;
+    volatile MUXSDKPlayerState _state;
```
nidhik, line 50: wasn't this already included in a previous PR?

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -77,16 +81,20 @@ - (void)attachAVPlayer:(AVPlayer *)player {
...
-                                                         [weakSelf updateLastPlayheadTime];
-                                                     }];
+            if([weakSelf isAdPlaying]) {
+                return;
+            } else if ([weakSelf isTryingToPlay]) {
+                [weakSelf startBuffering];
+            } else if ([weakSelf isBuffering]) {
+                [weakSelf dispatchPlaying];
+            } else {
+                [weakSelf dispatchTimeUpdateEvent:time];
```
nidhik, line 91: what does this do? Does it matter that we are early returning before calling `dispatchTimeUpdateEvent:`? I think you are right that we want to suppress the normal working of our SDK while the player being used to play an ad, but just curious

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -335,15 +343,17 @@ - (void)detachAVPlayer {
...
-    if (!_automaticVideoChange && !_didTriggerManualVideoChange) {
+    if ((!_automaticVideoChange && !_didTriggerManualVideoChange) || _isAdPlaying) {
         return;
     }
     if (_playerItem) {
         if (_didTriggerManualVideoChange) {
             _didTriggerManualVideoChange = false;
         }
         [self stopMonitoringAVPlayerItem];
+
```
nidhik, line 354: extra ws

nidhik (commented):
Overall lgtm, just a few questions

nidhik (commented):
This might be far too much work, so let me know if it is- but it would be nice to have a UI test that tests the ads + Picture in Picture mode which would get run by the buildkite pipeline

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.h
StephanieZuniga: yeah I need to update this PR with the lastest release. I opened it before merging those.

StephanieZuniga:
> This might be far too much work, so let me know if it is- but it would be nice to have a UI test that tests the ads + Picture in Picture mode which would get run by the buildkite pipeline

Should be pretty straight forward. I'll add it.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
StephanieZuniga: Yeah actually I intentionally return before dispatching time event cause otherwise it updates the playhead with the ad playhead value. At least that's the behavior I was seeing, does it make sense? Not sure what the server does with this event.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
StephanieZuniga: I thought it provided better readability. Can remove if you disagree.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
nidhik: I went digging and it is [used by the Mux Core internal heartbeat tracker ](https://github.com/muxinc/mux-stats-sdk-objc/blob/c10563f8933d2dfd7c2d1e74724c4af00abb699f/MuxCore/MuxCore/Trackers/MUXSDKInternalHeartbeatTracker.m#L83)to decide whether or not to fire a `hb` event. 

It is not sent by the server because it is not an external event. The events that are external (and therefore processed by the server) you can find here: https://github.com/muxinc/mux-stats-sdk-objc/blob/master/MuxCore/MuxCore/Trackers/MUXSDKExternalEventTracker.m#L43

nidhik: I think therefore the effect of this is that we will be suppressing the normal function of the internal heartbeat tracker and it won't fire any internal heartbeats at all until we are no longer playing the ad. 

This does have the effect that playhead time will not get updated. That's because we are essentially disabling this: https://github.com/muxinc/mux-stats-sdk-objc/blob/c10563f8933d2dfd7c2d1e74724c4af00abb699f/MuxCore/MuxCore/Trackers/MUXSDKPlaybackTimeTracker.m#L69

But there are other trackers like the rebuffer tracker which also look at internal heartbeats - do you know what effect this change would have on those trackers? https://github.com/muxinc/mux-stats-sdk-objc/blob/master/MuxCore/MuxCore/Trackers/MUXSDKRebufferTracker.m

I used the UI test case you added (thank you!) to test out the rebuffering metrics for PiP views and it kind of looks like this change makes us miss rebuffer events sometimes. Could you please take a look and lmk what you think? I described it here: 
https://app.shortcut.com/ios-sdks/story/10473/avplayer-sdk-splits-ads-and-content-into-different-views-when-picture-in-picture-is-enabled

nidhik: We found that it does take some time for AVPlayer to resume playing the content but that the notification for when an ad is done playing happens right away. Full details here, but it does not seem possible/feasible to track the time update events between when the ad stopped playing to when the player resumes content. https://app.shortcut.com/ios-sdks/story/10473/avplayer-sdk-splits-ads-and-content-into-different-views-when-picture-in-picture-is-enabled

nidhik (approved):

nidhik (approved):

PR #127, opened 2021-12-15
Release/2.8.0

StephanieZuniga:


apps/video-demo/video-demo/VideoPlayerController.swift
```diff
@@ -42,6 +42,22 @@ class VideoPlayerController: AVPlayerViewController, IMAAdsLoaderDelegate, IMAAd
         let playerBinding = MUXSDKStats.monitorAVPlayerViewController(self, withPlayerName: playName, customerData: customerData)
         imaListener = MuxImaListener.init(playerBinding: playerBinding!)
         player!.play()
+        
+//        self.testSeekingProgramatically()
+    }
+    
+    func testSeekingProgramatically() {
```
nidhik, line 49: Ah I missed this in the original PR. Thanks for adding to the video-demo app. Could we also add a test case to the Demo app for programmatic seeks @StephanieZuniga? So that this programmatic seeks are tested by the buildkite pipeline

nidhik (approved):

apps/video-demo/video-demo/VideoPlayerController.swift
StephanieZuniga: hmmm sure but should I do it on a separate branch taken out of this release or just for a separate release? Where should we do it? Also, the pipeline would just be running it cause I dont think we have a way of actually checking the events except how we manually check for the other UITests, or am I misunderstanding?

nidhik: Just for a separate release. Yes, the pipeline would just generate those views but I find it helpful to be able to sanity check the views that are generated before releasing

StephanieZuniga: sounds good I will finish this release then and create a ticket to track that test and make sure we don't forget

PR #131, opened 2022-01-03
Clear customer metadata stored under `playerName` when `destroyPlayer` is called

nidhik:
## Shortcut
https://app.shortcut.com/ios-sdks/story/10841/clear-customer-metadata-stored-under-playername-when-destroy-player-is-called

## Overview

We need to make sure to clear out all customer metadata when a player monitor is destroyed so that the same player name may be reused.


StephanieZuniga:
Seems like the unit tests are failing.

StephanieZuniga (approved):
Changes look good except there seems to be an issue with the unit tests

PR #133, opened 2022-01-04
Release v2.9.0

StephanieZuniga:
* Fix programmatic seek latency and sequence issues
* Add Carthage binary project specification
* Clear customer metadata stored under `playerName` when `destroyPlayer` is called
* Add device detection info new properties for overrides support
* Update release documentation


MUXSDKStats/Podfile
```diff
@@ -5,20 +5,23 @@ target 'MUXSDKStats' do
   # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
   # use_frameworks!
   # Pods for MUXSDKStats
-  pod 'Mux-Stats-Core', '~>3.6'
+  #  pod 'Mux-Stats-Core', '~>3.6'
+  pod 'Mux-Stats-Core', :git => 'https://github.com/muxinc/stats-sdk-objc.git', :branch => 'release/3.8.0'
```
nidhik, line 9: I think we need to update this podfile and the podfiles for all the demo apps to point to the right version of Mux Core

nidhik (changes requested):

MUXSDKStats/Podfile
StephanieZuniga: Updated

nidhik (approved):

PR #134, opened 2022-01-06
Programmatic seek fix

StephanieZuniga:
Listen to player property timeControlStatus to register seeked event after seeking


apps/DemoApp/DemoApp/ViewController.m
```diff
@@ -318,7 +318,16 @@ - (void)playerItemDidReachEnd:(NSNotification *)notification {
     MUXSDKCustomerVideoData *videoData = [MUXSDKCustomerVideoData new];
     videoData.videoTitle = @"Apple Keynote";
     videoData.videoId = @"applekeynote2010";
-    [MUXSDKStats videoChangeForPlayer:DEMO_PLAYER_NAME withVideoData:videoData];
+    
+    MUXSDKCustomerData *customerData = [
```
StephanieZuniga, line 322: This is unrelated but it was to get rid of the notification for using the deprecated videoChange function

XCFramework/MUXSDKStats.xcframework/Info.plist
```diff
@@ -6,18 +6,16 @@
 	<array>
 		<dict>
 			<key>LibraryIdentifier</key>
-			<string>tvos-arm64_x86_64-simulator</string>
+			<string>ios-arm64_armv7</string>
```
audrarodriguez, line 9: Is it ok to have this changes? 🤔

StephanieZuniga: tbh we always get them automatically haha and I'm not sure if there's a way to avoid them it might also be cause the xcode version I am using to generate the xcframework is different than the one in buildkite which I used to generate the last xcframework in the release branch. 

But yeah it doesn't hurt.

nidhik (approved):

PR #136, opened 2022-01-07
Release v2.10.0

StephanieZuniga:
- Fix programmatic seek events for iOS 15.0
- Fix picture in picture ads compatibility with mux-stats-google-ima 0.7.0

Authored-by: Stephanie Zuniga <szuniga@konradgroup.com>


audrarodriguez (approved):

apps/video-demo/Podfile.lock
```diff
@@ -1,6 +1,6 @@
 PODS:
   - GoogleAds-IMA-iOS-SDK (3.14.4)
-  - Mux-Stats-AVPlayer (2.9.0):
+  - Mux-Stats-AVPlayer (2.10.0):
     - Mux-Stats-Core (~> 3.8)
   - Mux-Stats-Core (3.8.0)
   - Mux-Stats-Google-IMA (0.6.0):
```
nidhik, line 6: shouldn't this be 0.7.0?

nidhik (changes requested):
Don't we also need to update the individual Podfiles and podspec?

apps/video-demo/Podfile.lock
nidhik: How are the PiP views generated in CI looking? I'm assuming it's not working? Maybe we should add a step to check those to the README since it's totally a manual sanity check right now :)

nidhik: I assume this means the PiP views in the CI environment aren't working? We should probably add a step to check those before releasing in the README, since it is just a manual sanity check right :)

StephanieZuniga: the ima sdk 0.7.0 release has a dependency on this sdk so in order to make that release we need to first get this release done. That is why we had to extract the examples logic that would have required ima sdk 0.7.0. On the next release we can add the example and update the podfile locks to target the new ima version

nidhik: ohh right I remember this discussion now - thanks

nidhik (approved):

PR #138, opened 2022-01-20
Add GitHub Actions workflow to build and test

englishm-mux:
Hack week work on iOS CI with @nidhik.

### DIFF
--- a/apps/DemoApp/run-tests.sh
+++ b/apps/DemoApp/run-tests.sh
@@ -6,6 +6,13 @@ rm -Rf XCFramework
 # reset simulators
 xcrun -v simctl shutdown all
 xcrun -v simctl erase all
+
+# Fetch artifact if running via buildkite (GitHub Actions fetches the artifact in a prior step)
+if command -v buildkit-agent > /dev/null 2>&1;
+then
+    buildkite-agent artifact download "MUXSDKStats.xcframework.zip" . --step ".buildkite/build.sh"
+fi
+
 unzip MUXSDKStats.xcframework.zip
 cd apps/DemoApp
 pod deintegrate && pod update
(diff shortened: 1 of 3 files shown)


apps/DemoApp/run-tests.sh
```diff
@@ -6,7 +6,13 @@ rm -Rf XCFramework
 # reset simulators
 xcrun -v simctl shutdown all
 xcrun -v simctl erase all
-buildkite-agent artifact download "MUXSDKStats.xcframework.zip" . --step ".buildkite/build.sh"
+
+# Fetch artifact if running via buildkite (GitHub Actions fetches the artifact in a prior step)
+if command -v buildkit-agent > /dev/null 2>&1;
```
nidhik, line 11: oh nice, so we can support buildkite still if needed. Thanks for adding this

nidhik (approved):
🎉  looks awesome

apps/DemoApp/run-tests.sh
englishm-mux: Oops! It looks like I may have gotten something wrong with my bash conditional here, based on this recent buildkite failure:

> `unzip:  cannot find or open MUXSDKStats.xcframework.zip, MUXSDKStats.xcframework.zip.zip or MUXSDKStats.xcframework.zip.ZIP`

https://buildkite.com/mux/stats-sdk-avplayer/builds/483#7f30c47a-b87b-48cd-8cb2-d7611b451029

englishm-mux: Oh, I see it now. `s/buildkit/buildkite/` 🤦‍♂️

nidhik: Ah good catch. I actually just deactivated the webhook for the Buildkite pipeline but can reactivate it if you want to test

englishm-mux: I did open [a PR](https://github.com/muxinc/mux-stats-sdk-avplayer/pull/140) to fix, but I think this really only affects my pride. 😆

PR #139, opened 2022-01-20
remove branch specific trigger

nidhik:


.github/workflows/build-and-test.yml
```diff
@@ -2,9 +2,7 @@ name: Build
 
 on:
   push:
-    branches: [ master ]
```
englishm-mux, line 5: Removes `branches` [filter](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-filters) from the `push` event trigger.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

Will now trigger any time

> One or more commits are pushed to a repository branch or tag.

https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push

englishm-mux: (This sounds like the desired behavior)

.github/workflows/build-and-test.yml
```diff
@@ -2,9 +2,7 @@ name: Build
...
   pull_request:
-    branches: [ master ]
```
englishm-mux, line 7: ~~Removes `branches` [filter](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-filters) from `pull_request` event trigger.~~

~~https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request~~

> ~~Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.~~

englishm-mux: ~~This may be redundant if we're already running the workflow on every push.~~

englishm-mux: Addressed by https://github.com/muxinc/mux-stats-sdk-avplayer/pull/139/commits/8e486129422285ebb45fb7c452ce6381fed6f1d2 which removes the `pull_request` event trigger altogether.

englishm-mux (approved):
Looks good to me! 👍

PR #143, opened 2022-02-10
Fix: Update rendition change

audrarodriguez:
Updated code to handle rendition change to only dispatch when playback started

### DIFF
--- a/MUXSDKStats/MUXSDKStats.xcodeproj/xcshareddata/xcschemes/MUXSDKStats.xcscheme
+++ b/MUXSDKStats/MUXSDKStats.xcodeproj/xcshareddata/xcschemes/MUXSDKStats.xcscheme
@@ -26,7 +26,13 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "NO">
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "TEST"
+            isEnabled = "YES">
+         </CommandLineArgument>
+      </CommandLineArguments>
       <Testables>
          <TestableReference
             skipped = "NO">
--- a/MUXSDKStats/MUXSDKStats.xcodeproj/xcshareddata/xcschemes/MUXSDKStats.xcscheme
+++ b/MUXSDKStats/MUXSDKStats.xcodeproj/xcshareddata/xcschemes/MUXSDKStats.xcscheme
@@ -26,13 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "NO">
-      <CommandLineArguments>
-         <CommandLineArgument
-            argument = "TEST"
-            isEnabled = "YES">
-         </CommandLineArgument>
-      </CommandLineArguments>
+      shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
          <TestableReference
             skipped = "NO">
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -504,9 +504,6 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
             }
         }
     }
-    if ([self IsTesting]) {
-        _started = YES;
-    }
     
     if (videoDataUpdated) {
         MUXSDKVideoData *videoData = [[MUXSDKVideoData alloc] init];
@@ -669,10 +666,6 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
         return NO;
     }
     return YES;
-}
-
--(BOOL)IsTesting {
-    return [[NSProcessInfo processInfo].arguments containsObject:@"TEST"];
 }
 
 - (BOOL)isPlaying {
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -194,7 +194,7 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
 - (void) handleRenditionChangeInAccessLog:(AVPlayerItemAccessLog *) log {
     AVPlayerItemAccessLogEvent *lastEvent = log.events.lastObject;
     float advertisedBitrate = lastEvent.indicatedBitrate;
-    if (advertisedBitrate == 0 && !_started) {
+    if (advertisedBitrate != 0 && !_started) {
         _lastAdvertisedBitrate = advertisedBitrate;
         return;
     }
--- a/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -194,19 +194,20 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
 - (void) handleRenditionChangeInAccessLog:(AVPlayerItemAccessLog *) log {
     AVPlayerItemAccessLogEvent *lastEvent = log.events.lastObject;
     float advertisedBitrate = lastEvent.indicatedBitrate;
-    BOOL bitrateHasChanged = ![self doubleValueIsEqual:@(_lastAdvertisedBitrate) toOther:@(advertisedBitrate)];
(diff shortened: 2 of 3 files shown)


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -194,20 +194,18 @@ - (void) handleRenditionChange:(NSNotification *) notif {
...
-        if(isStartingPlayback) {
-            // This is not a renditionchange but the player playing the first rendition.
-            _lastAdvertisedBitrate = advertisedBitrate;
-            return;
-        }
-        NSLog(@"MUXSDK-INFO - Switch advertised bitrate from: %f to: %f", _lastAdvertisedBitrate, advertisedBitrate);
-        [[NSNotificationCenter defaultCenter] postNotificationName:RenditionChangeNotification object: @{
-            RenditionChangeNotificationInfoAdvertisedBitrate: @(advertisedBitrate)
-        }];
+    if (advertisedBitrate <= 0 && !_started) {
```
StephanieZuniga, line 197: could this have a negative value? Shouldn't it just be the same comparison we were doing for the first: `[self doubleValueIsEqual:@(_lastAdvertisedBitrate) toOther:@(0)];` If we are doing early exit shouldn't it be different than instead of less than? I don't think this can take negative values.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -522,8 +524,9 @@ - (void)checkVideoData {
         if (_videoURL) {
             [videoData setVideoSourceUrl:_videoURL];
         }
-        if (_lastAdvertisedBitrate > 0) {
+        if (_lastAdvertisedBitrate > 0 && _started) {
             [videoData setVideoSourceAdvertisedBitrate:@(_lastAdvertisedBitrate)];
+            _lastDispatchedAdvertisedBitrate = _lastAdvertisedBitrate;
```
StephanieZuniga, line 529: Why did we move this?

StephanieZuniga (changes requested):

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
audrarodriguez: I moved this back, this needs to be done in the code where we check if those two are the same value to avoid doing unnecessary events.

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
audrarodriguez: Yes I change it to ==, so if the value is 0 and the playback hasn't started we assing the first advertisedBitrate and the early exit can happen

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -194,20 +194,18 @@ - (void) handleRenditionChange:(NSNotification *) notif {
...
-        if(isStartingPlayback) {
-            // This is not a renditionchange but the player playing the first rendition.
-            _lastAdvertisedBitrate = advertisedBitrate;
-            return;
-        }
-        NSLog(@"MUXSDK-INFO - Switch advertised bitrate from: %f to: %f", _lastAdvertisedBitrate, advertisedBitrate);
-        [[NSNotificationCenter defaultCenter] postNotificationName:RenditionChangeNotification object: @{
-            RenditionChangeNotificationInfoAdvertisedBitrate: @(advertisedBitrate)
-        }];
+    if (advertisedBitrate == 0 && !_started) {
```
StephanieZuniga, line 197: this condition should be different than 0, otherwise it means it would only assign this first _lastAdvertisedBitrate = 0.

StephanieZuniga (approved):

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -194,20 +194,18 @@ - (void) handleRenditionChange:(NSNotification *) notif {
 - (void) handleRenditionChangeInAccessLog:(AVPlayerItemAccessLog *) log {
     AVPlayerItemAccessLogEvent *lastEvent = log.events.lastObject;
     float advertisedBitrate = lastEvent.indicatedBitrate;
-    BOOL bitrateHasChanged = ![self doubleValueIsEqual:@(_lastAdvertisedBitrate) toOther:@(advertisedBitrate)];
```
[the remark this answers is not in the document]
audrarodriguez: I was confused with the check on the other function where we dispatch the rendition change event but this one is different so I put it back!

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -194,20 +194,18 @@ - (void) handleRenditionChange:(NSNotification *) notif {
...
-        if(isStartingPlayback) {
-            // This is not a renditionchange but the player playing the first rendition.
-            _lastAdvertisedBitrate = advertisedBitrate;
-            return;
-        }
-        NSLog(@"MUXSDK-INFO - Switch advertised bitrate from: %f to: %f", _lastAdvertisedBitrate, advertisedBitrate);
-        [[NSNotificationCenter defaultCenter] postNotificationName:RenditionChangeNotification object: @{
-            RenditionChangeNotificationInfoAdvertisedBitrate: @(advertisedBitrate)
-        }];
+    if (advertisedBitrate != 0 && !_started) {
```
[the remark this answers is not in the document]
audrarodriguez: Updated!

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -502,6 +504,10 @@ - (void)checkVideoData {
             }
         }
     }
+    if ([self IsTesting]) {
```
nidhik, line 507: I'm not sure about adding this type of testing only code into our SDK - can you give me some context here?

MUXSDKStats/MUXSDKStats.xcodeproj/xcshareddata/xcschemes/MUXSDKStats.xcscheme
```diff
@@ -26,7 +26,13 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "NO">
+      <CommandLineArguments>
```
nidhik, line 30: did you mean to commit this

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
audrarodriguez: This will only execute when the tests are running, the Test scheme has an argument with the flag "TEST" that will be recognized here: 
`-(BOOL)IsTesting {
    return [[NSProcessInfo processInfo].arguments containsObject:@"TEST"];
}`
This is because with the change of logic the renditionChange tests were failing because we are checking if playback has started, with this flag we can change the state of "_started" to true to fulfill the check I mentioned before but the change of state is only happening when we are running the unit tests because the argument was just added in that scheme.
This same approach we have it in Kaltura, we use the same argument on the tests to assign the mockedDispatcher or the normal dispatcher.

MUXSDKStats/MUXSDKStatsTests/MUXSDKStatsTests.m
```diff
@@ -843,19 +850,19 @@ - (void) testOrientationAndRenditionChangeEventSequence {
 
     [self assertPlayer:playName dispatchedDataEvents:@{
         @(1): [NSNull null],
-        @(4): @{BANDWIDTH: @(258157)},
-        @(6): @{BANDWIDTH: @(1927853)},
-        @(9): @{BANDWIDTH: @(258157)},
+        @(5): @{BANDWIDTH: @(258157)},
```
nidhik, line 853: how come we are still sending the same renditions even though now we are waiting for playback to start in order to send a renditionchange?

StephanieZuniga: I think essentially this is because of how the test works, the test forces the rendition change notifications so the change is always registered cause the playbackStartDate logic happens before this. So not on testing but on an actual flow if there is a rendition change that we get in the accessLog we check for playbackStartDate and then depending on that value we post a notification with the rendition change or just return earlier without posting the notification. However, these tests are not mocking the accessLog but instead directly forcing the rendition change notifications.

StephanieZuniga (approved):

nidhik (approved):

MUXSDKStats/MUXSDKStatsTests/MUXSDKStatsTests.m
audrarodriguez: Just wanted to confirm here. We're forcing the `RenditionChangeNotification` with valid data so as Steph said the tests are working with that notification instead of the accessLog, with the access log there's more logic to check to see if we can send or not this rendition notification but in this case the tests will just sent the notification with the data we want to test and verify.

PR #148, opened 2022-03-17
Fix crash in `AVMetadataItem` inspection

ejensen:
These changes prevent a crash caused by casting `[AVMetadataItem key]` to an `NSString` without checking the type first.

`[AVMetadataItem key]` is defined as `id<NSObject, NSCopying>` since it can be a type other than `NSString`. 

I'm witnessing crashes in v1.12.0 due to `[AVMetadataItem key]` returning an `NSNumber` and the line `[keyString hasPrefix:MUXSessionDataPrefix]` throws an `unrecognized selector sent to instance` exception.

These changes check that  `[AVMetadataItem key]` is an `NSString` before casting to it.


nidhik (approved):
Thank you for catching this!

PR #150, opened 2022-04-08
BeaconCollectionDomain

StephanieZuniga:
Deprecate old custom beacon implementation and add new beaconCollectionDomain that does not append the key.


MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -152,13 +152,65 @@ + (MUXSDKPlayerBinding *_Nullable)monitorAVPlayerViewController:(nonnull AVPlaye
...
                                          automaticErrorTracking:(BOOL)automaticErrorTracking
-                                                   beaconDomain:(nullable NSString *)domain {
+                                         beaconCollectionDomain:(nullable NSString *)collectionDomain {
 
     MUXSDKCustomerViewerData *viewerData = [customerData customerViewerData];
     if (viewerData != nil) {
         _customerViewerData = viewerData;
     }
 
+    [self initSDK];
```
nidhik, line 162: this is basically the same as the legacy/deprecated implementation right? Except for this one line
```
[MUXSDKCore setBeaconDomain:domain forPlayer:name];
```
Wondering if it's worth factoring the shared stuff out in its own method?

StephanieZuniga: I guess it depends on how soon are we gonna be removing the code. Since we chat about the mejor version on sprint planning my thoughts were this is gonna be removed soon and if I extract it into it's own function/method to avoid logic repetition, in this case it should be a couple more methods to preserve the sequence of things. And they would not really be methods that would be used elsewhere. So I thought if I leave it like this we can just go in remove the deprecated method and then we're done with it. It just feels like the use of any extra encapsulation of logic would be until we remove this from the code. If you think that's gonna take too long then it might be worth extracting into separate functions.

StephanieZuniga: Updated as discussed.

apps/DemoApp/DemoApp/ViewController.m
```diff
@@ -283,7 +283,7 @@ - (void)setupAVPlayerViewController:(AVPlayer *)player {
                                                                                      viewData:viewData
                                                                                    customData:customData
                                                                                    viewerData:viewerData];
-    _playerBinding = [MUXSDKStats monitorAVPlayerViewController:_avplayerController withPlayerName:DEMO_PLAYER_NAME customerData:customerData];
+    _playerBinding = [MUXSDKStats monitorAVPlayerViewController:_avplayerController withPlayerName:DEMO_PLAYER_NAME customerData:customerData automaticErrorTracking:YES beaconCollectionDomain:@"myDomain.com"];
```
nidhik, line 286: we should remove this bc it isn't expected to work right?

nidhik (approved):
lgtm, pending updates to DemoApp

apps/DemoApp/DemoApp/ViewController.m
StephanieZuniga: yep, my bad. Updated

PR #152, opened 2022-04-29
Improve scripts to fail for xcodebuild issues

StephanieZuniga:


nidhik (commented):
the script updates look good but I din't think we need to rebuild the framework

nidhik (approved):

PR #161, opened 2022-08-30
Fix Error Metadata being reported, but no Error Events

daytime-em:


nidhik (approved):

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -1014,7 +1014,7 @@ - (BOOL)isPlayerInErrorState {
     if (!_player || !_playerItem) {
         return NO;
     }
-    return _player.status == AVPlayerStatusFailed || _playerItem.status == AVPlayerItemStatusFailed;
+    return _player.error || _playerItem.error;
```
nidhik, line 1017: do we know that these are always fata? Because if they aren't we don't want to track it as an error. That is my one concern with this change

nidhik: it sounds like it should be only populated if it is fatal?
>By default, this value is nil. If a player reaches a [AVPlayer.Status.failed](https://developer.apple.com/documentation/avfoundation/avplayer/status/failed), the system populates this value with an error that describes the failure.

daytime-em: That is a really good question. I think I have to research, but we might be able to be a little more selective.

PR #164, opened 2022-10-17
Allow customers to override device metadata with their own values

daytime-em:
This should be a draft until the objc changes are released and I can revert the `Podfile` changes. ~Tests also need to be added~


MUXSDKStats/MUXSDKStats/MUXSDKStats+Internal.h
```diff
@@ -0,0 +1,23 @@
...
+//  Created by Emily Dixon on 10/17/22.
+//  Copyright © 2022 Mux, Inc. All rights reserved.
+//
+
+#ifndef MUXSDKStats_PropertyGetters_h
+#define MUXSDKStats_PropertyGetters_h
+
+#import "MUXSDKStats.h"
+
+@interface MUXSDKStats()
```
nidhik, line 16: Alright so what I remember about categories is that we used to use them but they don't get linked up correctly when building the static framework. It ends up crashing at runtime. 

I can't remember all access to the class is affected or just if the method on the category is called - perhas worth testing? And also if we do use this, can we please add a warning comment to this category warning of the pitfalls of using it outside of tests? It makes me a bit nervous to include it

PR #166, opened 2022-10-19
Update to Core 4.x + Remove Deprecated Buildkite Test Pipeline + XCode 14 / iOS16

daytime-em:


nidhik (approved):
Looks okay pending we bump the version number of AVPlayer as well - I think it would also require a major version rev. And the upgrading of Core also requires updating the min deployment version

daytime-em:
Note: Due to dependency and ci issues we must hold off on updating avplayer to 3.0.0 until the release is ready

PR #168, opened 2022-10-28
Add A Mac Catalyst Demo

daytime-em:
It's not that different from our existing demo but it *does* use SwiftUI and it allows us to test this issues like [this one in objc core](https://github.com/muxinc/stats-sdk-objc/issues/95) again in the future


daytime-em:
This includes the version bump for 3.0 and should remain a draft until 3.0 is released

apps/CatalystApp/CatalystApp/ContentView.swift
```diff
@@ -0,0 +1,59 @@
...
+
+import SwiftUI
+import AVKit
+import AVFoundation
+import MUXSDKStats
+
+struct ContentView: View {
+    
+    private static let MEDIA_URL = "https://stream.mux.com/v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM.m3u8"
+    private static let MUX_DATA_ENV_KEY = "jp06n8nvbl6u8g3u40cpvi9u2" // TODO: Your Key Here
```
nidhik, line 16: Let's replace with `YOUR_ENV_KEY_HERE`?

apps/CatalystApp/CatalystApp/AVPlayerViewSwiftUi.swift
```diff
@@ -0,0 +1,47 @@
...
+//  Created by Emily Dixon on 10/28/22.
+//
+import Foundation
+import SwiftUI
+import AVKit
+
+/**
+ * Wraps an `AVPlayerViewController` with SwiftUI, providing accwss to its `AVPlayer`
+ */
+struct AVPlayerViewSwiftUi : UIViewControllerRepresentable {
```
nidhik, line 14: This is the tiniest nit 😅 But since this is also an example app, I think `PlayerView` might be clearer? It's clearly using SwiftUI & AVKit so feel like this doesn't need to be in the name, plus I thought SwiftUI components tend to be named more generally. I also have no production experience with SwiftUI so please take this with a grain of salt

nidhik (approved):
looks good with minor comments

apps/CatalystApp/CatalystApp/AVPlayerViewSwiftUi.swift
daytime-em: I do like that way better!

PR #178, opened 2023-02-04
Add an explicit statically-linked target for building our static lib

daytime-em:
Splits our build targets into statically-linked and dynamically-linked libraries. This allows us to make the Cocoapods version of this SDK a dynamic library, a requirement for successfully linking with the latest tvOS. Following [this article](https://medium.com/microsoft-mobile-engineering/mixing-static-and-dynamic-linking-in-cocoapods-83b6d4252c59), we create both a static target and a dynamic target. Our build scripts choose the static target when building the static lib and the dynamic target when building the dynamic lib.

I don't think this will affect most peoples' build, but if it does we can also offer the explicitly statically-linked version as a pod.


andrewjl-mux (approved):

daytime-em:
This doesn't actually work. Let's release without it and see if this winds up being required

PR #187, opened 2023-08-14
Add state check when dispatching viewend

andrewjl-mux:
* Log warning when attempting viewend

* Adjust unit test logic to destroy player before checking event sequence

* Include viewend as expected event during testing


daytime-em (approved):
looks entirely reasonable to me!

PR #214, opened 2023-10-26
Releases/v3.4.2

andrewjl-mux:
## Improvements

* Pin MuxCore to specific version consistent across Cocoapods and Swift Package Manager


Package.swift
```diff
@@ -15,7 +15,7 @@ let package = Package(
         .package(
             name: "MuxCore",
             url: "https://github.com/muxinc/stats-sdk-objc.git",
-            from: "4.5.2"
+            revision: "982feb93c4601684b5e3eef9ad1a036e3ec117c7"
```
andrewjl-mux, line 18: Likewise to Cocoapods, pinning to a specific MuxCore version here.

scripts/version-check.sh
```diff
@@ -38,6 +38,19 @@ else
     exit 1
 fi
 
+echo "▸ Validating ${CARTHAGE_JSON_SPECIFICATION} version-specific path"
```
andrewjl-mux, line 41: This almost tripped me up just now, so added a check

daytime-em: Awesome!!

daytime-em (approved):

PR #216, opened 2023-10-30
chore: remove deprecated interfaces

andrewjl-mux:
This PR also converts the inline API docs to the docc format. Down the road we should be able to generate static documentation using the Xcode documentation compiler.


andrewjl-mux:
> lgtm, just make sure it's in a major release version

Yep! This was stacked on another PR but it's ultimately getting merged into releases/v4.0.0, which is a major bump from the current version (v3.4.2)

PR #217, opened 2023-10-30
feat: expose new monitoring API

andrewjl-mux:
Much of the new API spelling is currently WIP and newly added methods are largely unimplemented.


Sources/MUXSDKStatsObjc/MUXSDKStats.h
```diff
@@ -321,12 +321,12 @@ FOUNDATION_EXPORT
 ///
 /// Use this method to signal that the player is now playing
 /// a new video. The player name provided must been passed
-/// as the name in a monitorPlayer:withPlayerName:andConfig:
-/// call. The config provided should match the specifications
-/// in the Mux docs at https://docs.mux.com and should include
```
andrewjl-mux, line 326: @jsanford8 was `and should include all desired keys, not just those keys that are specific to this video.` intended to mean values like custom environment key should be set on the new `MUXSDKCustomerData`, if they were set previously and are expected to continue being used? Want to make sure I'm interpreting the doc right.

Sources/MUXSDKStatsObjc/MUXSDKMonitor.h
```diff
@@ -0,0 +1,397 @@
...
+///   - customerData: A MUXSDKCustomerData object with player,
+///   video, and view metadata
+- (void)updatePlayerName:(nonnull NSString *)playerName
+        withCustomerData:(nullable MUXSDKCustomerData *)customerData;
+
+#pragma mark - Orientation Change
+
+// TODO: Move MUXSDKViewOrientation to its own header
+
+// TODO: We should probably allow a starting orientation to be set when calling startMonitoring on an AVPlayer
```
andrewjl-mux, line 370: @jsanford8 what do you think? When `startMonitoring` is called for an `AVPlayer` I don't think there is a way for us to detect its orientation. (I need to confirm this.)

Sources/MUXSDKStatsObjc/MUXSDKMonitor.h
```diff
@@ -0,0 +1,397 @@
...
+/// Removes any observers monitoring the instance associated
+/// with the provided player binding
+///
+/// - Parameter playerBinding: The player binding for the
+/// instance to stop monitoring
+- (void)stopMonitoringWithPlayerBinding:(nonnull MUXSDKPlayerBinding *)playerBinding;
+
+#pragma mark - Automatic Video Change
+
+// TODO: Consider whether to use a single method with BOOL parameter vs separate methods here. Should this also be exposed as a parameter for the startMonitoring methods?
```
andrewjl-mux, line 295: I went back and forth on the BOOL parameter vs separate methods, anyone have strong opinions?

Sources/MUXSDKStatsObjc/MUXSDKMonitor.h
jsanford8: Yes, that would probably be a good thing. The other option (what people do today) is fire an orientationchange event right away to signal the first orientation, which should work.

Sources/MUXSDKStatsObjc/MUXSDKMonitor.h
```diff
@@ -0,0 +1,397 @@
...
+      withPlayerName:(nonnull NSString *)playerName
+     fixedPlayerSize:(CGSize)fixedPlayerSize;
+
+#pragma mark - Stop Monitoring
+
+/// Removes any observers monitoring the instance with the
+/// provided player name
+///
+/// - Parameter playerName: The player name to stop monitoring
+- (void)stopMonitoringWithPlayerName:(nonnull NSString *)playerName;
```
andrewjl-mux, line 283: Objective-C lacks type-based overloading (unlike Swift) so the `WithPlayerName:` is required here unless `stopMonitoringWithPlayerBinding:` also gets dropped.

Sources/MUXSDKStatsObjc/MUXSDKMonitor.h
andrewjl-mux: Combined these into one method.

Sources/MUXSDKStatsObjc/MUXSDKMonitor.h
andrewjl-mux: Yeah I think the other option sounds like the way to go for now, to avoid multiplying the different permutations of parameters on these `startMonitoring` methods. Sending an orientation change right after calling `startMonitoring` should still work the same way it did before.

At some point I think it would make sense to either add something similar to `MUXSDKCustomerData` but for parameters specific to this layer or extend `MUXSDKCustomerData` to include these additional parameters.

cjpillsbury (approved):
🚀

PR #219, opened 2023-11-03
fix: use right BOOL type

andrewjl-mux:
Changes `MUXSDKStats` to use the conventionally standard Objective-C `BOOL` type.


daytime-em (approved):
`YES` it looks pretty good

PR #247, opened 2024-08-14
fix: improve AVPlayer currentItem transition handling

andrewjl-mux:
- fix: dispatch view end when AVPlayerItem changes
- fix: skip dispatching view init when player has no player item


daytime-em (approved):

cjpillsbury (approved):
Curious if there are any breaking changes as a result of this, but feels like the "right" default behavior regardless. 👍

andrewjl-mux:
Before this change we'd dispatch a `viewinit` event whenever [`AVPlayer` `currentItem`](https://developer.apple.com/documentation/avfoundation/avplayer/1387569-currentitem) changed. This fix also dispatches `viewend` before that, which is closer to how we work on Android and in `mux-embed`. 

This _un_-breaks the edge case where `destroyPlayer` is called at the same time as when these other transitions are happening.

PR #249, opened 2024-10-11
fix: tear down view on hard application termination

andrewjl-mux:
* dispatch a viewend automatically and tear down player monitoring when the application shuts down
* remove commented out log lines


cjpillsbury (approved):
😎

PR #250, opened 2024-10-16
build: add xcconfig, to simplify version updates

andrewjl-mux:
- [x] Add xcconfig for each target, to simplify version updates
- [x] Remove iOS 11 deployment target flags that aren't applicable (and would result in build failures on Xcode 15+ if they were)
- [x] Add Xcode path log to static build script for consistency
- [x] Skip SPM and Cocoapods tests on non-release branches


daytime-em (approved):
my only regret is that i have but one ✅ to give

PR #258, opened 2024-12-13
update: allow starting new views with `videoChange` without changing AVPlayerItem

daytime-em:


MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -1232,6 +1247,9 @@ - (BOOL) doubleValueIsEqual:(NSNumber *) x toOther:(NSNumber *) n {
 }
 
 - (void)didTriggerManualVideoChange {
+    // TODO: Should we deprecate? we no longer need this method. videoChange can happen whenever it's needed now, not just when AVPlayerItem changes
```
andrewjl-mux, line 1250: Strong +1 to deprecate

andrewjl-mux: If keeping this functionality around makes sense we could also expose a better named API with the same functionality while deprecating this one. (And we'd drop the deprecated method next major version bump.)

andrewjl-mux (approved):

MUXSDKStats/MUXSDKStats/MUXSDKStats.m
```diff
@@ -603,13 +603,10 @@ + (void)videoChangeForPlayer:(nonnull NSString *)name withCustomerData:(nullable
     MUXSDKCustomerVideoData *videoData = [customerData customerVideoData];
     MUXSDKCustomData *customData = [customerData customData];
     
-    if (!(videoData || viewData || customData)) {
```
andrewjl-mux, line 606: Does removing this change anything?

daytime-em: This would prevent the `dispatchViewEnd` and `dispatchVideoChange` calls from going out, so there'd be no video change. I don't think that makes sense to do from an API standpoint.

The customer doesn't necessarily have to change any of their customer data when they do `videochange`. They may choose to do so later on in the new view via `setCustomerData`, may only provide info that wouldn't need to change, or may choose not to add any additional info at all.

andrewjl-mux: Agree that the updated behavior makes more sense. We should probably call this out in the release notes so that anyone relying on how this works is aware of the change.

andrewjl-mux: (And thanks for the explainer!)

MUXSDKStats/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -403,38 +403,50 @@ - (void)detachAVPlayer {
...
 }
 
+- (void)dispatchVideoChange {
+    // end the current view if it's not already been ended
+    if (_state != MUXSDKPlayerStateViewEnd) {
+        [self dispatchViewEnd];
+    }
+    
+    [self.playDispatchDelegate videoChangedForPlayer:_name];
+    // TODO: test that we have all the metadata we need on the subsequent view. On android we needed to catch up with the current state
```
andrewjl-mux, line 413: Does this refer to external stuff like `MUXSDKCustomerData` or something else?

MUXSDKStats/MUXSDKStats/MUXSDKStats.h
```diff
@@ -257,7 +257,7 @@ FOUNDATION_EXPORT
 
  */
 + (void)videoChangeForPlayer:(nonnull NSString *)name
-            withCustomerData:(nullable MUXSDKCustomerData *)customerData;
+            withCustomerData:(nonnull MUXSDKCustomerData *)customerData;
```
andrewjl-mux, line 260: This is a breaking change and requires a major version bump. Particularly for Swift callers.

I'm not sure if it's absolutely required too implement the required functionality in this PR, though. Based on everything else here I don't feel like it is?

andrewjl-mux (changes requested):
Latest SDK changes look good overall! 

If we absolutely need the API break for `videoChangeForPlayer: withCustomerData:` we need to take a major version bump to release this.

MUXSDKStats/MUXSDKStats/MUXSDKStats.h
daytime-em: Good point. I don't need to make this nullable.

PR #265, opened 2025-03-13
fix: Missing Source Resolution (and therefore WAB and VQ) in some cases

daytime-em:
When we can't get a source resolution or source bitrate, we can't calculate WAB, Scaling, or other VQ-related metrics.

## Cases where Src Res/Bitrate was not picked up

* Late registration: If an `AVPlayerItem` or `AVPlayer` had been used before (as when preloading or when pooling/reusing players), we won't see the events that report things like Source URL, Source Resolution, and Advertised Bitrate
* No Bitrate Advertised: If a media item doesn't advertise a bitrate, VQ and WAB can't be calculated. But we can calculate an effective bitrate based on the data we have. Examples of these cases: HLS streams that are just a bare MP (as discussed recently), HLS streams whose MVP doesn't have any `EXT-X-BITRATE` tags (they are optional), people playing progressive mp4 files (possibly; I didn't test this)

## Future-facing concerns

* It would be really great to be able to test these cases using some kind of modern, easy test harness app that allows us to make assertions in automated tests... 🤔 
* If we were listening for KVO updates on `AVPlayerItem.tracks` (including the `.initial` flag), we might be able to do some of these calculations there instead, and there'd be fewer paths that lead to this logic. Unsure if this is a real improvement or not, since the way this is written is a little more clear: "When I monitor an AVPlayerItem, I need to check for these specific things".
  * I see that as a stylistic choice, and I'm interested in others' opinions about it, if any

## Draft Until
This PR should stay a draft until I have made some automated tests to cover this case, and removed the test code from `BasicPlaybackExampleViewController`

### DIFF
--- a/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -603,35 +603,15 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
         if (lastEvent) {
             long long transferredBytes = lastEvent.numberOfBytesTransferred;
             NSTimeInterval downloadedDuration = lastEvent.segmentsDownloadedDuration;
-            NSLog(@"findMostRecentAdvertisedBitrateForPlayerItem: last event has transferred bytes %lld", transferredBytes);
-            NSLog(@"findMostRecentAdvertisedBitrateForPlayerItem: last event has duration %f", downloadedDuration);
-//            NSNumber *transferredBits = [NSNumber numberWithFloat: transferredBytes * 8];
             long long transferredBits = transferredBytes * 8;
             double calculatedBitrate = transferredBits / downloadedDuration;
-            NSLog(@"findMostRecentAdvertisedBitrateForPlayerItem: calculated bitrate %f", calculatedBitrate);
-            NSLog(@"findMostRecentAdvertisedBitrateForPlayerItem: calculated bitrate as long %ld", (long)calculatedBitrate);
-            return calculatedBitrate;
+            return (long)calculatedBitrate; // any rounding is fine
         } else {
-            NSLog(@"findMostRecentAdvertisedBitrateForPlayerItem: last event was nil");
+            NSLog(@"findMostRecentAdvertisedBitrateForPlayerItem: no events");
         }
-        return -1;
-
-//        // TODO: This only works if there is a bitrate indicated in the MVP/Media Playlist
-//        //  Do we even want to use this one still
-//        for (int i = 0; i < events.count; i++) {
-//            AVPlayerItemAccessLogEvent *ev = events[i];
-//            double bitrateFromLog = ev.indicatedBitrate;
-//            if (bitrateFromLog > 0) {
-//                recentBitrateEvent = ev;
-//            }
-//        }
     }
     
-//    if (recentBitrateEvent) {
-//        return recentBitrateEvent.indicatedBitrate;
-//    } else {
-//        return -1;
-//    }
+    return -1;
 }
 
 - (nullable NSString *)getURLStringIfAvailableOnAVPlayerItem:(nonnull AVPlayerItem *)item {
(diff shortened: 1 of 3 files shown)


Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -539,6 +593,67 @@ - (CGSize)getSourceDimensions {
     return CGSizeMake(0, 0);
 }
 
+- (long)findMostRecentAdvertisedBitrateForPlayerItem:(nonnull AVPlayerItem *)item {
+    AVPlayerItemAccessLog *accessLog = [item accessLog];
+    
+    if (accessLog && accessLog.events) {
+        NSArray<AVPlayerItemAccessLogEvent *> *events = [accessLog.events copy];
+
```
daytime-em, line 601: Something else we can do is try to scan this log for valid `indicatedBitrate`s, though this doesn't always work

Maybe the calculation strategy below could be a fallback instead of the default behavior.. Curious what we think here, I'm pretty comfy with either option

Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -539,6 +593,67 @@ - (CGSize)getSourceDimensions {
...
+                    AVMediaType mediaType = assetTrack.mediaType;
+                    if (mediaType && [mediaType isEqualToString:AVMediaTypeVideo]) {
+                        return playerItemTrack;
+                    }
+                }
+            }
+        }
+//    } @catch (NSException *exception) {
+//        NSLog(@"%@", exception.reason);
+//    }
```
CFKevinRef, line 652: Added a note to the other try/catch, but here it's still possible (if extremely unlikely) that the `tracks` property could change out from under you. In any case, `for in` loops are easier to write:
```suggestion
    for (AVPlayerItemTrack *playerItemTrack in item.tracks) {
        if ([playerItemTrack.assetTrack.mediaType isEqual:AVMediaTypeVideo]) {
            return playerItemTrack;
        }
    }
```
I also suggested here to use `isEqual:` instead of `isEqualToString:`. The only "benefit" of `isEqualToString:` is that it throws an exception if the argument or receiver is not a string. So it's almost always a good idea to just use `isEqual:`.

Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -518,6 +571,7 @@ - (CGRect)getViewBounds {
 }
 
 - (CGSize)getSourceDimensions {
+    // TODO: What throws here? Is this try-catch even necessary
```
CFKevinRef, line 574: Probably since it manually iterates over `_player.currentItem.tracks` which means any of `_player`, `currentItem`, or `tracks` could all change mid-iteration. Using "fast iteration" takes care of this - behind the scenes it grabs an `objectEnumerator` property off the collection one time and uses that to implement the `for in` loop. `enumerateObjectsUsingBlock:` has the same benefit if you still need the index for some reason.
In any case I've flagged this for replacement in another ticket.

Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -440,6 +441,58 @@ - (void)monitorAVPlayerItem {
...
+            
+            long alreadyAdvertisedBitrate = [self findMostRecentAdvertisedBitrateForPlayerItem:_playerItem];
+            if (alreadyAdvertisedBitrate > 0) {
+                _lastAdvertisedBitrate = alreadyAdvertisedBitrate;
+                _lastDispatchedAdvertisedBitrate = alreadyAdvertisedBitrate;
+                videoData.videoSourceAdvertisedBitrate = [
+                    NSNumber numberWithFloat:alreadyAdvertisedBitrate
+                ];
+            }
+            [MUXSDKCore dispatchEvent:dataEvent forPlayer:[self name]];
```
CFKevinRef, line 493: It seems like this (and possibly the surrounding code) duplicates what's already in `checkVideoData`, maybe there's a way to avoid that?

CFKevinRef (commented):
One suggestion for the try/catch which I had already kind of thought through in setting up the cleanup tickets. Otherwise I think if we can find a way to avoid duplication this would be an even more manageable change!

PR #266, opened 2025-03-14
Add test fixtures and consistent code signing

CFKevinRef:
### **Integration Test Workspace** added at `Fixtures/IntegrationTests`
- All tests (included imported unit tests from the base SDK) are repeated with sanitizers enabled - adding extra runtime checks for undefined behavior, bad address references, and threading.
- Uses an XCWorkspace and a symlink up to the main package, which allows us to import the unit tests and run them on device (not possible with a bare package)
- Adds a multiplatform "Integration Test Host" app target which pulls in the main package. Currently does very little, but we can build scenarios here and it will run like a real app.
- Adds an "Integration Tests" target where we write our tests - this can access APIs from the test host directly (including specific views/scenarios we've built and of course our SDK)
- Adds a "UI Tests" target. The difference here is that there is no API access to the host app, and any test cases must be implemented via UI automation. This type of testing will be especially useful in our example apps, but here we still can use it to test any UI-level interactions and accessibility.

### Code Signing
- Using an Apple official sample code style "disambiguator" pattern for both tests and example apps. We use automatic code signing, but don't set the development team explicitly anywhere:
- Examples and test harnesses run it on simulator as-is
- For-device building/running works by selecting a development team from Xcode's drop-down menu. The first time you try to build/run for device, Xcode gives you an error which automatically takes you here.
- A special git-ignored local file (`Configuration/CodeSigning.local.xcconfig`) can be used to set this preference without impacting git history or hard-coding anything
- CI uses its own config file here (`Configuration/CodeSigning.mux.xcconfig`) which also works for anyone on the Mux team - just make a symlink from the hidden local one to this one like the CI scripts do.

### Continuous Integration
  - Tests are built for all supported platforms (including simulator and device) and for each type of sanitizer.
  - Tests are run for all simulator platforms, exporting a merged XCResult bundle with the results as an artifact. This can be fed to further tooling or opened in Xcode to inspect as if the tests were just run locally
  - Test products bundles are exported for all physical device platforms. These are code-signed, self-contained bundles that can be run directly on a developer device or packaged up for Sauce Labs. (Note that the above simulator tests are also built to this format, there's nothing special about the build process here)


daytime-em (approved):
nice!!

PR #267, opened 2025-03-14
Use standard Examples/ directory

CFKevinRef:
`Examples` is the standard location for example apps in a Swift Package repository layout. Broken out into this PR for clarity as it's mostly moved files after #266


daytime-em (approved):
🎉

PR #268, opened 2025-04-09
Cleanup: remove private ivars from public headers

CFKevinRef:
This makes for a much cleaner public interface with fewer visible implementation details.


daytime-em (approved):
LGTM! ✨ 👍

PR #269, opened 2025-04-10
Swift rendition tracker WIP

CFKevinRef:
Work in progress for rendition tracking in Swift, converted to work mostly via Combine


daytime-em:
> combine
Very cool!

Love the changes so far! Thanks for the preview

Sources/MUXSDKStatsInternal/Publishers+AVPlayerItem.swift
```diff
@@ -0,0 +1,315 @@
...
+            }
+            .switchToLatest()
+            .removeDuplicates()
+            .share()
+
+        let activeVariantPublisher = variantsFuture?
+            .replaceError(with: [])
+            .combineLatest(naturalSizePublisher)
+            .map { variants, naturalSize in
+                variants.first { $0.videoAttributes?.presentationSize == naturalSize }
```
daytime-em, line 185: If a customer's rendition ladder has variants at different bitrates or codecs but at the same resolution, will this still be accurate?

CFKevinRef: There's a limitation we can do without some really hairy access log matching logic. I'm adding add a check to make sure we at least don't update with the _wrong_ bitrate, and a basic look into the access log to match on the most recent bitrate requested.

But yes, `.first` really isn't a good choice here.

Sources/MUXSDKStatsInternal/MUXSDKVideoData.swift
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        // TODO: videoAttributes.codecTypes
+    }
+
+    @available(iOS 15, tvOS 15, *)
+    func updateWithRenditionInfo(track: AVAssetTrack, on playerItem: AVPlayerItem) async {
+        // This is less exact and only used after all other variant matching options are exhausted
+        // Still obtain it immediately in case another variant switch happens
+        // TODO: add timeout here?
```
CFKevinRef, line 44: hmm, yes this needs a timeout

Sources/MUXSDKStatsInternal/MUXSDKVideoData.swift
```diff
@@ -0,0 +1,222 @@
...
+        async let bitratesFromAccessLog = playerItem.indicatedBitratesInAccessLog
```
CFKevinRef, line 45: hmm, yes this needs a timeout 😅

Sources/MUXSDKStatsInternal/MUXSDKVideoData.swift
```diff
@@ -0,0 +1,214 @@
...
+            updateWithPresentationSize(videoAttributes.presentationSize)
+        }
+
+        // TODO: videoAttributes.codecTypes
+    }
+
+    @available(iOS 15, tvOS 15, *)
+    func updateWithRenditionInfo(track: AVAssetTrack, on playerItem: AVPlayerItem) async {
+        // This is less exact and only used after all other variant matching options are exhausted
+        // Still obtain it immediately in case another variant switch happens
```
daytime-em, line 35: I'm not sure if I understand this comment. Is it about `indicatedBitratesInAccessLog` or is it about the overall usage of this method?

I see that we use `updateWithRenditionInfo` for getting our initial rendition, and when we don't have AVMetrics available, which seems like an interpretation of the comment? But I also see that we only use `bitratesFromAccessLog` conditionally, which also seems like an interpretation of the comment 😅 .

Sources/MUXSDKStatsInternal/MUXSDKVideoData.swift
```diff
@@ -0,0 +1,214 @@
...
+        // Matching below based on the following assumptions:
+        //
+        //  HLS Attribute     | AVAssetVariant key path
+        //  ------------------+---------------------------------
+        //  RESOLUTION        | videoAttributes.presentationSize
+        //  FRAME-RATE        | videoAttributes.nominalFrameRate
+        //  CODECS            | (audio/video)Attributes.codecTypes
+        //  VIDEO-RANGE       | videoAttributes.videoRange
+        //  BANDWIDTH         | peakBitRate
+        //  AVERAGE-BANDWIDTH | averageBitRate
```
daytime-em, line 96: This table is extremely helpful! 🙌

Sources/MUXSDKStatsInternal/AVPlayerItem+AccessLog.swift
```diff
@@ -0,0 +1,44 @@
+import AVFoundation
+
+@available(iOS 15, tvOS 15, *)
+extension AVPlayerItem {
+    nonisolated var indicatedBitratesInAccessLog: some BidirectionalCollection<Double> {
```
daytime-em, line 5: Making sure I understand: 

If we have indicated bitrates in the access log already we return (that one's easy)

If we need to wait for some, how does this property behave? Does it wait for the first indicated bitrate and return that?

daytime-em (approved):
This looks great!! Very excited

I had some comments, but they were mostly questions. Nothing that I would want to block merging on

Sources/MUXSDKStatsInternal/AVPlayerItem+AccessLog.swift
daytime-em: ... your new commit makes this much more clear.

PR #274, opened 2025-04-23
Fix framework-style header includes

CFKevinRef:
Objective-C projects including SDK version 4.6.0 via Swift Package Manager need to use quoted includes:
```objc
#import "MUXSDKStats.h"
```
This reverts to the previous behavior:
```objc
#import <MUXSDKStats/MUXSDKStats.h>
```


daytime-em (approved):
Looks reasonable to me

PR #277, opened 2025-05-13
VoD Playback Tests

Fabrizio-rar:
This PR adds integration tests for VoD playback:

In MUXSDKCore Mock:

- Created a getEventsForPlayer() method that returns all of the MUXSDKBaseEvent for a given playerId
- Added a timeStamps and timeDeltas dicts, in order to store the timestamps and timedeltas for players
- Modified the mock_dispatchEvent() logic so that if an event is a MUXSDKTimeUpdateEvent, its data is added into the timeStamps and timeDeltas dict
- Created getTimeStampsForPlayer and getTimeDeltasForPlayer to give access to the dicts

In IntegrationTests:

- Created a bridging header file in the IntegrationTests project to be able to use the MUXSDKCore Mock implementation
- Implemented atomic methods with the idea of reusability in additional playback tests apart from VoD playback
- Implemented the VoD Playback Tests


Fixtures/IntegrationTests/Mocks/MUXSDKCore+Mock.h
```diff
@@ -0,0 +1,27 @@
+//
+//  MUXSDKCore+Mock.h
+//  MUXSDKStatsTests
+//
+//  Created by Nidhi Kulkarni on 2/4/20.
+//  Copyright © 2020 Mux, Inc. All rights reserved.
```
esbie, line 6: Let's not give Nidhi toooo much credit here :D

Fixtures/IntegrationTests/Mocks/MUXSDKCore+Mock.h
```diff
@@ -0,0 +1,27 @@
...
++ (void) swizzleDispatchEvents;
++ (void) resetCapturedEvents;
++ (id<MUXSDKEventTyping>) eventAtIndex:(NSUInteger) index forPlayer:(NSString *)playerId;
++ (NSUInteger) eventsCountForPlayer:(NSString *)playerId;
++ (MUXSDKDataEvent *) globalEventAtIndex:(NSUInteger)index;
++ (NSUInteger) globalEventsCount;
++ (NSArray<MUXSDKDataEvent *> *) snapshotOfGlobalEvents;
++ (NSArray<MUXSDKDataEvent *> *) snapshotOfEventsForPlayer:(NSString *)playerId;
++ (NSArray<MUXSDKBaseEvent *> *) getEventsForPlayer: (NSString *)playerId;
++ (NSArray *)getTimeStampsForPlayer:(NSString *)playerId;
```
daytime-em, line 23: If these are counting playhead deltas, I think we need to name them this way. Like `getPlayheadTimestampsForPlayer` or something.

We also count 'real'/'wall-clock' time for some of our metrics (eg, rebuffering time or seek latency). We are probably going to want to test our actual metric values at some point, so this code may evenutally need to refer to both playhead and wall-clock timestamps unambiguously.

daytime-em (changes requested):
Looks good to me, except I wanted to insist on a naming change.

Fixtures/IntegrationTests/Mocks/MUXSDKCore+Mock.h
Fabrizio-rar: changed it!

CFKevinRef (approved):
Copying the mocks is probably the easiest solution here, so I think this is good, especially without requiring big structural changes or overly clever hacks. Thanks!

PR #280, opened 2025-05-28
Crash When Calling AVPlayer play() off of Main Thread

fabrizio-rar:
This PR fixes a crash when calling AVPlayer `play` method when off the main application thread, instead now it logs a warning.

The underlying issue was the call to `getViewBounds` in the `MUXSDKAVPlayerViewControllerBinding` implementation, which accesses UI logic, operation that cannot be done from outside the main thread.

A test was created in order to ensure that this crash is avoided and the method instead of giving out an error, it returns an empty CGRect.

### DIFF
--- a/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -1375,10 +1375,6 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
 #endif
 }
 
-- (CGRect)getViewBounds {
-    return [self getViewBoundsValue].CGRectValue;
-}
-
 - (nullable NSValue *)getViewBoundsValue {
     if (![NSThread isMainThread]) {
         return nil;
@@ -1459,10 +1455,6 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
     return [_view videoRect];
 }
 
-- (CGRect)getViewBounds {
-    return [self getViewBoundsValue].CGRectValue;
-}
-
 - (nullable NSValue *)getViewBoundsValue {
     return [NSValue valueWithCGRect:_view.bounds];
 }
@@ -1508,10 +1500,6 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
                       0.0,
                       0.0
                       );
-}
-
-- (CGRect)getViewBounds {
-    return [self getViewBoundsValue].CGRectValue;
 }
 
 - (nullable NSValue *)getViewBoundsValue {


spuppo-mux (approved):
Not sure if buildConfiguration change was intentional in IntegrationTests scheme (Fixtures/IntegrationTests/IntegrationTests.xcworkspace/xcshareddata/xcschemes/IntegrationTestHost.xcscheme)

Otherwise LGTM.

spuppo-mux (approved):

Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -1292,6 +1292,10 @@ - (CGRect)getVideoBounds {
 }
 
 - (CGRect)getViewBounds {
+    if (![NSThread isMainThread]) {
+        NSLog(@"MUXSDK-WARNING - getViewBounds called from a background thread.");
+        return CGRectZero;
```
CFKevinRef, line 1297: We would instead return `nil` here...

CFKevinRef (changes requested):
I'm thinking we should not send zero player size when this happens, but instead just send nothing (by not setting player width/height). I'm concerned that we'll be sending random zero values whenever we're off the main thread.

We change `- (CGRect)getViewBounds` to `- (nullable NSValue *)getViewBounds` everywhere, and check for `nil` before updating player data with [`CGRectValue`](https://developer.apple.com/documentation/foundation/nsvalue/cgrectvalue?language=objc).

Maybe zero is never correct and we should just check for `CGRectIsNull` instead, but using `NSValue` and checking for `nil` seems like the safest and clearest, while also not changing any other behavior.

Sources/MUXSDKStats/include/MUXSDKStats/MUXSDKPlayerBinding.h
```diff
@@ -39,7 +39,7 @@ typedef NS_ENUM(NSUInteger, MUXSDKViewOrientation) {
 - (void)detachAVPlayer;
 - (void)programChangedForPlayer;
 - (void)prepareForAvQueuePlayerNextItem;
-- (CGRect)getViewBounds;
+- (nullable NSValue *)getViewBounds;
```
CFKevinRef, line 42: Ah I see this is technically a public header, hmm. I'm sure no one is relying on it, but let's be safe:
- change the new method to `getViewBoundsValue`
- keep `getViewBounds` but mark it deprecated with `DEPRECATED_ATTRIBUTE`
- implement `getViewBounds` in the base class with something like `[self getViewBoundsValue].CGRectValue` so it technically still works.

CFKevinRef (changes requested):
Cool this should be much safer. Just a couple of suggestions

PR #281, opened 2025-06-03
Mini cleanup of mocks/tests

CFKevinRef:
Some quick fixes to make testing more reliable:
- synchronizes all access to the global static event dicts that are used during testing
- use unique IDs for player names
- enable strict concurrency checking
- adds some nullability annotations
- downloads any artifacts generated by Sauce Labs and adds them to the Build Kite artifacts

...Also restricts Sauce Labs builds to the only currently working device: iPhone 16e


Fixtures/IntegrationTests/IntegrationTests/IntegrationTests.swift
```diff
@@ -118,10 +106,13 @@ struct IntegrationTests {
     }
     
     @Test func vodPlaybackTest() throws {
-        let playerName = "vodPlayerName"
+        let playerName = "vodPlayer \(UUID().uuidString)"
         MUXSDKCore.swizzleDispatchEvents()
         MUXSDKCore.resetCapturedEvents(forPlayer: playerName)
```
spuppo-mux, line 111:
```suggestion
```

spuppo-mux (approved):
Was working on this too. Your changes look cleaner though!

Fixtures/IntegrationTests/IntegrationTests/IntegrationTests.swift
spuppo-mux: (It doesnt appear on the suggestion, but this is called again one line below)

CFKevinRef: Thanks I'll add this to the next commit

PR #284, opened 2025-06-05
bug fix - Load Values Asynchronously For Keys causes stall

spuppo-mux:
This PR tries to fix a customer reported issue causing main thread to stall (for more than 2 seconds).
 
[From the Apple docs](https://developer.apple.com/documentation/avfoundation/avasynchronouskeyvalueloading/loadvaluesasynchronously(forkeys:completionhandler:)?language=objc)
> Regardless of the number of keys specified, the system calls the completion handler only once per invocation of this method. The system calls this method:
> 
> Synchronously if the asset already loaded the specified keys, or if an I/O error or other format-related error occurs immediately.
> 
> Asynchronously when the values of the specified keys become loaded, if a loading error occurs at a later stage of processing, or you cancel loading. The system calls the closure on a background queue, so you should dispatch control back to the main queue before performing any user interface-related operations.

I couldn't reproduce this issue yet, but I added error checking on the completion handler and moved the execution off the main thread.

Added note: this function appears as deprecated on Apple Docs, however this is for Swift only
Screenshot 2025-06-05 at 6 04 54 PM

Screenshot 2025-06-05 at 6 06 12 PM

Screenshot 2025-06-05 at 6 06 07 PM

### DIFF
--- a/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -493,7 +493,7 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
 			[asset statusOfValueForKey:@"metadata"
 								 error:&error];
 			if (status != AVKeyValueStatusLoaded || error != nil) {
-                // NSLog(@"Failed to load metadata: %@", error);
+				NSLog(@"MUXSDK-ERROR - Mux failed to load asset metadata for player name: %@", _name);
                 return;
             }
 
--- a/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -488,12 +488,17 @@ NSString * RemoveObserverExceptionName = @"NSRangeException";
     dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^{
         [asset loadValuesAsynchronouslyForKeys:@[ @"metadata" ]
                              completionHandler:^{
+            __typeof(weakSelf) strongSelf = weakSelf;
+            if (strongSelf == nil) {
+                return;
+            }
+
             NSError *error = nil;
             AVKeyValueStatus status = [asset statusOfValueForKey:@"metadata"
                                                            error:&error];
             if (status != AVKeyValueStatusLoaded || error != nil) {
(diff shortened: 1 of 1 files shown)


Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -484,35 +484,53 @@ - (void)dispatchSessionData {
...
-            MUXSDKSessionDataEvent *dataEvent = [MUXSDKSessionDataEvent new];
-            [dataEvent setSessionData: sessionData];
-            [MUXSDKCore dispatchEvent:dataEvent 
-                            forPlayer:[strongSelf name]];
-        }
-    }];
+            
+            if ([sessionData count] > 0) {
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    __typeof(weakSelf) strongSelf = weakSelf;
```
CFKevinRef, line 521:
```suggestion
                    __strong __typeof(weakSelf) strongSelf = weakSelf;
```

Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -484,35 +484,53 @@ - (void)dispatchSessionData {
     AVAsset *asset = _player.currentItem.asset;
     // Load Session Data from HLS manifest
     __weak MUXSDKPlayerBinding *weakSelf = self;
+    
+    dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^{
+        [asset loadValuesAsynchronouslyForKeys:@[ @"metadata" ]
+                             completionHandler:^{
+            __typeof(weakSelf) strongSelf = weakSelf;
```
CFKevinRef, line 491:
```suggestion
            __strong __typeof(weakSelf) strongSelf = weakSelf;
```
Ha, looks like this was a bug in the original implementation, let's go ahead and fix it now too (I saw your commit avoiding the `_name` access below, nice)

CFKevinRef (approved):
This looks good!. These minimal changes should _at least_ improve the situation to a background thread hang instead of main thread, but I have a feeling it will resolve them completely. That and I'm always happy to get things off the main thread!

Sources/MUXSDKStats/MUXSDKPlayerBinding.m
spuppo-mux: Ah, now I wonder if this was causing issues

PR #285, opened 2025-06-09
Strict Swift Concurrency

CFKevinRef:
Updates package to reflect Xcode 16 requirements (swift-tools-version:6.0), enables strict concurrency checking, and makes related updates to comply with this strict compiler mode.

No unsafe or escape hatch solutions were used except for getting around the unnecessarily strict sendability requirement on `MainActor.assumeIsolated`, and comments were added to explain.

Tested locally as well with Xcode 16.0, which is the oldest version that can make App Store submissions.

### DIFF
--- a/Sources/MUXSDKStatsInternal/Future+async.swift
+++ b/Sources/MUXSDKStatsInternal/Future+async.swift
@@ -2,9 +2,10 @@ import Combine
 
 @available(iOS 13, tvOS 13, *)
 extension Future where Failure == Never {
-    convenience init(priority: TaskPriority? = nil, operation: sending @escaping @isolated(any) () async -> Output) {
+    convenience init(isolation: isolated (any Actor)? = #isolation, operation: @escaping () async -> sending Output) {
         self.init { promise in
-            Task(priority: priority) {
+            Task {
+                _ = isolation
                 promise(.success(await operation()))
             }
         }
--- a/Sources/MUXSDKStatsInternal/MUXSDKVideoData.swift
+++ b/Sources/MUXSDKStatsInternal/MUXSDKVideoData.swift
@@ -36,13 +36,20 @@ extension MUXSDKVideoData {
     }
 
     @available(iOS 15, tvOS 15, *)
-    func updateWithRenditionInfo(track: AVAssetTrack, on playerItem: AVPlayerItem) async {
+    static func renditionInfo(track: AVAssetTrack, on playerItem: AVPlayerItem) async -> sending Self {
+        let videoData = Self()
         // Obtain recent bitrate stats immediately in case another variant switch happens
-        async let bitratesFromAccessLog = withTimeout(of: 5.0) {
-            try await playerItem.indicatedBitratesInAccessLog
+        async let bitratesFromAccessLog = withTimeout(seconds: 5) {
+            Array(try await playerItem.indicatedBitratesInAccessLog.suffix(3))
         }
         // Start loading this early:
-        async let assetVariantsOrNil = (playerItem.asset as? AVURLAsset)?.load(.variants)
+        async let assetVariantsOrNil = {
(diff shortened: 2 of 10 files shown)


Sources/MUXSDKStatsInternal/Task+Timeout.swift
```diff
@@ -4,18 +4,26 @@ struct TimeoutError: Error {
...
+                }
+                group.addTask {
+                    try await workTask.value
+                }
+                group.addTask {
+                    try await Task.sleep(nanoseconds: UInt64(seconds * Double(NSEC_PER_SEC)))
+                    throw TimeoutError()
+                }
+                let firstResult = await group.nextResult()!
+                continuation.resume(with: firstResult)
```
CFKevinRef, line 25: `withCheckedThrowingContinuation` can now return immediately. `withThrowingTaskGroup` always waits until all subtasks return, which means waiting until they successfully notice cancellation and clean up. See tests added for blocking operations and never-returning operations!

daytime-em (approved):
Very cool! LGTM!

Sources/MUXSDKStatsInternal/Future+async.swift
```diff
@@ -2,9 +2,10 @@ import Combine
 
 @available(iOS 13, tvOS 13, *)
 extension Future where Failure == Never {
-    convenience init(priority: TaskPriority? = nil, operation: sending @escaping @isolated(any) () async -> Output) {
+    convenience init(isolation: isolated (any Actor)? = #isolation, operation: @escaping () async -> sending Output) {
         self.init { promise in
-            Task(priority: priority) {
+            Task {
+                _ = isolation
```
daytime-em, line 8: Comprehension check (for me): Does referencing `isolation` here put the Task inside that context? Or is it for some other purpose?

And before this change, was the inside of this `Task {` in any isolation context at all? And like, same question for the inside of `operation`

Sources/MUXSDKStatsInternal/MUXSDKVideoData.swift
```diff
@@ -36,13 +36,20 @@ extension MUXSDKVideoData {
     }
 
     @available(iOS 15, tvOS 15, *)
-    func updateWithRenditionInfo(track: AVAssetTrack, on playerItem: AVPlayerItem) async {
+    static func renditionInfo(track: AVAssetTrack, on playerItem: AVPlayerItem) async -> sending Self {
```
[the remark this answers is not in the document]
daytime-em: tiny naming gripe: `MUXSDKVideoData.makeWithRenditionInfo(...)` since it's creating something

daytime-em (commented):
ok, i had some extra comments, but nothing major at all

Sources/MUXSDKStatsInternal/Future+async.swift
CFKevinRef: currently this is used in an ad-hoc async `map` function implemented via `flatMap` returning a `Future`, and once I fix capturing event timing earlier in the chain I'll clean this up:
```swift
// current:
.flatMap { videoAssetTrack in
    // capture timing immediately
    let timing = PlaybackEventTiming(playerItem: self)

    return Future {
        guard let videoAssetTrack else {
            return (timing, MUXSDKVideoData())
        }
        return (timing, await MUXSDKVideoData.renditionInfo(track: videoAssetTrack, on: self))
    }
}

// cleaned up, where `Future` is an implementation detail in this new `map()`
.map { videoAssetTrack in
    guard let videoAssetTrack else {
        return MUXSDKVideoData()
    }
    return await MUXSDKVideoData.renditionInfo(track: videoAssetTrack, on: self)
}
```

daytime-em: Thanks for the detailed explanation! I think I get it :D

PR #294, opened 2025-06-30
Backgrounding / Resuming a View Integration Test

ignasio-mux:
Integration test that analyzes the behavior of the SDK when the application is sent to background and resumed.

Test steps:

- Begin playback of content
- Wait 5 seconds
- Simulate backgrounding/resuming (5 seconds in background)
- Wait 5 second

The test will be successful if:
- There is a Pause event right after entering background
- There are no events during background
- There is a Play event right after entering foreground

This PR closes https://github.com/muxinc/psdks_and_dev_rel_team/issues/329

### DIFF
--- a/Fixtures/IntegrationTests/IntegrationTests/IntegrationTests.swift
+++ b/Fixtures/IntegrationTests/IntegrationTests/IntegrationTests.swift
@@ -119,8 +119,6 @@ struct IntegrationTests {
     }
     
     func assertBackground(backgroundSeconds: Double, with player: AVPlayer, for playerName: String) async {
-        NSLog("## Simulate app backgrounding for \(backgroundSeconds) seconds")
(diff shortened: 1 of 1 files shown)


CFKevinRef (changes requested):
I have a few minor suggestions for style just for future reference, and I appreciate the implementation using an `async` test too. Overall though this needs to be an XCUITest. It isn't really testing background/foreground behavior, it's posting a notification and manually pausing/resuming.
Unfortunately that means actually filling out the IntegrationTestHost app, which is currently just a stub, and writing our first UI test in that target too. So maybe let's put this one on hold?

CFKevinRef:
Doing some PR cleanup, going to go ahead and close this one as it's going to be implemented elsewhere.

PR #296, opened 2025-07-01
Fatal Errors Integration Test

ignasio-mux:
Integration test that analyzes the Error Events triggered when there is a fatal error in the playback of a video.

Test steps:
- Attempt to play a invalid URL
- Wait for the error to occur
- Check the events
- Exit the view

The test will be successful if:
- There is at least one Error Event
- The error severity is fatal
- There registered Error Events have an Error Code
- There registered Error Events have an Error Message


daytime-em (approved):
🙌

PR #303, opened 2025-07-23
add timestamp helpers and validity checks on conversion

CFKevinRef:
Guarantees that timestamp/duration values use milliseconds and avoids nonsense values such as NaN, infinities, undefined values from `CMTime`, etc.

Broken out from #291


esbie (approved):
seems reasonable. would be even better with 1 unit test

CFKevinRef:
> seems reasonable. would be even better with 1 unit test

Had to work around this bizarre bug in Xcode 16 causing it to crash when specifying tests a certain way, didn't want to push immediately as it actually required a force quit sometimes

PR #304, opened 2025-07-24
require unit tests to pass on CI before longer tasks

CFKevinRef:
With this change, longer running tasks (integration tests and sauce labs) will not start until the package builds and unit tests pass. Currently all of these tasks have the same priority, so PRs with failing unit tests will still eat up tons of CI time running the more involved tests.


esbie (approved):
good call

PR #306, opened 2025-07-30
Releases/v4.8.0

CFKevinRef:


Package.swift
```diff
@@ -4,12 +4,6 @@ import PackageDescription
 
 let package = Package(
     name: "MUXSDKStats",
-    platforms: [
-        .iOS(.v12),
-        .macCatalyst(.v13),
-        .tvOS(.v12),
-        .visionOS(.v1),
-    ],
```
CFKevinRef, line 12: These are the default minimum versions, so this block does nothing. It's considered best practice for a utility package like ours to omit these and use availability annotations on individual APIs.

Sources/MUXSDKStats/MUXSDKPlayerBinding.m
```diff
@@ -2,15 +2,7 @@
...
 
-#if MUX_COCOAPODS
-#if __has_include("MUXSDKStats-Swift.h")
-#import "MUXSDKStats-Swift.h"
-#else
-#import <MUXSDKStats/MUXSDKStats-Swift.h>
-#endif
-#else
 @import MUXSDKStatsInternal;
-#endif
```
CFKevinRef, line 13: CocoaPods is now built as another project that consumes our SPM library, so no more trying to keep both systems happy.

PR #316, opened 2025-08-14
Use Mux assets in tests

CFKevinRef:
Uses well-known Mux assets for integration tests, modifying tests for timing to work with the new assets.

Additionally integration tests were themselves crashing, leading to useless results. This PR also changes the force-unwrapping of optionals to throwing `#require`s, but with no other attempt to change behavior.


daytime-em:
Looks like they're failing on the simulator as well. Could be something in our actual tests

daytime-em (approved):
🙌

PR #326, opened 2025-09-19
Automatically follow the player property on AVPlayerViewController/AVPlayerLayer

CFKevinRef:
Automatically observes changes to [`AVPlayerViewController.player`](https://developer.apple.com/documentation/avkit/avplayerviewcontroller/player) and [`AVPlayerLayer.player`](https://developer.apple.com/documentation/avfoundation/avplayerlayer/player) to set up/tear down monitoring.

Previously most of the SDK's monitoring setup/update methods would fail and do nothing if the `player` property of a view controller or layer was `nil`.

Implemented using Swift's Key-Value Observing helpers for safety and convenience.

Also fixes the `updateAVPlayerViewController:...` and `updateAVPlayerLayer:` methods. Previously they did not actually update the view controller or layer, just checked the player associated with those objects. The original object was still retained by the binding.

### DIFF
--- a/Sources/MUXSDKStats/MUXSDKStats.m
+++ b/Sources/MUXSDKStats/MUXSDKStats.m
@@ -156,7 +156,7 @@ static MUXSDKCustomerCustomDataStore *_customerCustomDataStore;
     // Destroy any previously existing player with this name.
     [self destroyPlayer:name];
 
-    if (player.player) {
+    {
         MUXSDKCustomerPlayerData *playerData = customerData.customerPlayerData;
         MUXSDKCustomerVideoData *videoData = customerData.customerVideoData;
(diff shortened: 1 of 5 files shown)


Sources/MUXSDKStats/MUXSDKStats.m
```diff
@@ -156,7 +156,7 @@ + (MUXSDKPlayerBinding *_Nullable)monitorAVPlayerViewController:(nonnull AVPlaye
     // Destroy any previously existing player with this name.
     [self destroyPlayer:name];
 
-    if (player.player) {
+    {
```
CFKevinRef, line 159: kept this as the beginning of a code block to avoid whitespace churn

PR #329, opened 2025-09-23
Remove unused getVideoBounds from bindings

CFKevinRef:
This `getVideoBounds` is not used internally or exposed publicly.


daytime-em (approved):
-25 👍 🚢

PR #354, opened 2026-04-01
Implement text track changes

CFKevinRef:
Implements the `MUXSDKTextTrackChange` (`texttrackchange`) event using the `AVMediaSelection` APIs.
- A best effort is made to match these to the presented `AVAssetTrack`s in order to get format information.
- Adds some utilities which will be useful in detecting/inspecting audio tracks and improving rendition change events
- Actual feature implementation is confined to its own `TextTrackChangeEvents.swift`
- Debounce with configurable time duration (default 0.25s) avoids noisy bursts of media selection changes. This seems to happen especially when initializing a player, or when switching between subtitles and cc.

---

> [!NOTE]
> **Medium Risk**
> Adds new Combine/async-driven event generation for subtitle/closed-caption selection changes and wires it into `PlayerMonitor`, which can affect event volume/timing and relies on best-effort track/metadata matching.
> 
> **Overview**
> Implements `texttrackchange` reporting by introducing `TextTrackChangeEvents`, which listens for `AVPlayerItem.mediaSelectionDidChangeNotification`, captures the selected legible option (name/language/type), and emits `MUXSDKTextTrackChangeEvent`s with debouncing and duplicate suppression.
> 
> Adds best-effort matching of the selected media option to presented `AVAssetTrack`s to derive `textTrackFormat`, plus small AVFoundation utilities (async metadata/group loading), a lazy `playerData` timing helper on `MUXSDKPlaybackEvent`, and updates `PlayerMonitor` to merge the new text-track event stream. Includes new tests covering initial/changed events and the async `Publisher.map` helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63988f81b41f37d88d9dcfcb9e838bd4d5db652f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
```diff
@@ -0,0 +1,230 @@
...
+        } else {
+            self.init(
+                textTrackEnabled: false as NSNumber,
+                textTrackType: nil,
+                textTrackFormat: nil,
+                textTrackName: nil,
+                textTrackLanguage: nil)
+        }
+
+        playerData?.updateWithTiming(timing)
```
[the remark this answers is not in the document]
CFKevinRef: The MuxCore initializer creates this automatically, but this code could still be better

eugene-mux:
do we want to merge this into master, or to the feature branch?

Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
```diff
@@ -0,0 +1,230 @@
...
+
+            self.trackType = trackType
+            self.trackFormat = trackFormat
+        }
+    }
+
+    let playerItem: AVPlayerItem
+    let minimumIntervalBetweenEvents: TimeInterval
+
+    init(playerItem: AVPlayerItem, minimumIntervalBetweenEvents: TimeInterval = 0.25) {
```
eugene-mux, line 115: just curious, where did we get the default value of `0.25` from?

Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
```diff
@@ -0,0 +1,230 @@
...
+        case .subtitle:
+            self = .subtitles
+        case .closedCaption:
+            self = .closedCaptions
+        default:
+            return nil
+        }
+    }
+
+    @available(iOS 13, tvOS 13, *)
```
eugene-mux, line 17: Does it makes sense, for consistency reasons, to have this property available starting from os version `15`, as `TextTrackChangeEvents` itself (which it seems to be depended on) is available starting from `15`? Same question for `MUXSDKTextTrackFormat` extension. Am I missing something?

Sources/MUXSDKStatsInternal/Utilities/AVMediaSelection.swift
```diff
@@ -0,0 +1,64 @@
+import AVFoundation
+
+@available(iOS 15, tvOS 15, *)
+extension AVMediaSelection {
```
eugene-mux: what do you think about renaming this file to something like `AVMediaSelection+Extension` to clearly show in its filename that it is an extension?

Sources/MUXSDKStatsInternal/Utilities/AVMediaSelectionOption+Extensions.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+extension AVMetadataIdentifier {
+
+    static let hlsNameAttribute = AVMetadataIdentifier("m3u8/NAME")
+}
+
+@available(iOS 15, tvOS 15, *)
+extension AVMediaSelectionOption {
```
eugene-mux, line 9: how about moving this extension to its own file, what's your preference on it? I think we can use this pattern for all the extensions we introduce.

Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
```diff
@@ -0,0 +1,230 @@
...
+@preconcurrency import MuxCore
+
+extension MUXSDKTextTrackType {
+    init?(_ mediaType: AVMediaType) {
+        switch mediaType {
+        case .subtitle:
+            self = .subtitles
+        case .closedCaption:
+            self = .closedCaptions
+        default:
```
eugene-mux, line 12: we only map `.subtitle` and `.closedCaption` to `MUXSDKTextTrackType`. That means if AVFoundation reports the selected legible option as `.text`, `MediaSelectionOptionInfo` becomes `nil`, and we end up emitting `textTrackEnabled = false` even though a text track is actually selected. Since we already treat `.text` as subtitle traffic elsewhere, should we handle` .text `here as well so we don’t under-report enabled text tracks?

eugene-mux (changes requested):
Looks good, please check my questions and comments before we merge.

Also, do we want to add tests for this new text-track event path as part of this PR?

Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
CFKevinRef: I wasn't able to get AVPlayer to produce anything other than .subtitle or .caption but I don't see the harm in it, I'll make this a bit more permissive.

Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
CFKevinRef: These are based on the minimum OS versions that can use it. In this case, it's because `CMFormatDescription.MediaType` is available in iOS/tvOS 13. It just so happens that it's only called from higher version components, but still this annotation's job is to report capability.

Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
CFKevinRef: Just a small fraction of a second that worked in observation/testing, but left tunable and very visible.

Sources/MUXSDKStatsInternal/Features/TextTrackChangeEvents.swift
CFKevinRef: this extension remains unchanged, but any legible option is now supported in the implementation. We still won't send anything for this property since we don't know what it is.

eugene-mux (approved):

CFKevinRef:
⬆️ Just realized I had put the tests for that async map helper on the wrong branch, still have one more test suite coming for the `TextTrackChangeEvents` feature struct

eugene-mux:
> ⬆️ Just realized I had put the tests for that async map helper on the wrong branch, still have one more test suite coming for the `TextTrackChangeEvents` feature struct

Sounds good, thanks for adding those!

PR #364, opened 2026-06-16
NAT-458: Fix audio codec profile reporting

eugene-mux:
## Summary

Fix audio track codec reporting to prefer CoreMedia's richest decodable audio format before falling back to the format description subtype.

This preserves the existing fallback behavior, but lets AAC variants such as HE-AAC and HE-AACv2 report their more specific RFC 6381 codec strings instead of being flattened to AAC-LC.

## Root Cause

`CMFormatDescription.mediaSubType` can expose the most-compatible AAC subtype for HE-AAC/HE-AACv2 tracks, which reports `mp4a.40.2`. The audio format description can also include a richer decodable format, where HE-AACv2 maps to `kAudioFormatMPEG4AAC_HE_V2` / `mp4a.40.29`.

## Validation

- `xcodebuild -scheme MUXSDKStats -destination 'generic/platform=iOS Simulator' build`
- Manual local HE-AACv2 HLS playback verified in the Mux dashboard:
Screenshot 2026-06-16 at 12 51 52 PM

---

> [!NOTE]
> **Low Risk**
> Analytics-only metadata extraction for audio track change events; behavior is additive with an explicit fallback and new unit tests.
> 
> **Overview**
> **Fixes audio track codec strings in Mux stats** so HE-AAC, HE-AACv2, USAC, and similar profiles report their specific RFC 6381 values (e.g. `mp4a.40.29`) instead of being flattened to AAC-LC (`mp4a.40.2`) when `CMFormatDescription.mediaSubType` only exposes the most-compatible subtype.
> 
> `AssetTrackInfo.codecString(for: CMFormatDescription)` now consults **`CMAudioFormatDescriptionGetRichestDecodableFormat`** and maps `mFormatID` through the existing `AudioFormatID` → codec mapping, then **falls back** to the prior `mediaSubType` path when no richer format is available.
> 
> Adds parameterized tests over synthetic audio format descriptions plus a small **`CMAudioFormatDescriptionCreate`** test helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 439dfbe949b1610b1701294cfa22587ace949941. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/MUXSDKStatsInternal/Features/AudioTrackChangeEvents.swift
+++ b/Sources/MUXSDKStatsInternal/Features/AudioTrackChangeEvents.swift
@@ -211,6 +211,12 @@ struct AudioTrackChangeEvents {
                 return nil
             }
 
+            let audioFormatDescription = formatDescription as CMAudioFormatDescription
+            if let richestFormat = CMAudioFormatDescriptionGetRichestDecodableFormat(audioFormatDescription),
+               let codecString = codecString(for: richestFormat.pointee.mASBD.mFormatID) {
+                return codecString
+            }
+
             return formatDescription.mediaSubType.rawValue.rfc6381CodecsParameterValue
         }
 
--- a/Tests/MUXSDKStatsInternalTests/AudioTrackChangeEventsTests.swift
+++ b/Tests/MUXSDKStatsInternalTests/AudioTrackChangeEventsTests.swift
@@ -73,6 +73,27 @@ struct AudioTrackChangeEventsTests {
         #expect(AudioTrackChangeEvents.AssetTrackInfo.codecString(for: formatID) == expected)
     }
 
+    @Test(arguments: [
+        (AudioFormatID(kAudioFormatMPEG4AAC), "mp4a.40.2"),
+        (AudioFormatID(kAudioFormatMPEG4AAC_HE), "mp4a.40.5"),
+        (AudioFormatID(kAudioFormatMPEG4AAC_HE_V2), "mp4a.40.29"),
+        (AudioFormatID(kAudioFormatMPEGD_USAC), "mp4a.40.42"),
+        (AudioFormatID(kAudioFormatEnhancedAC3), "ec-3"),
+        (AudioFormatID(kAudioFormatAC3), "ac-3"),
+    ])
+    func codecStringForFormatDescription(formatID: AudioFormatID, expected: String?) throws {
+        guard #available(iOS 15, tvOS 15, *) else {
+            return
+        }
+
+        // Synthetic format descriptions validate this overload and its fallback behavior.
+        // AVFoundation-created HE-AAC descriptions are still needed to exercise mediaSubType
+        // flattening to AAC-LC while the richest decodable format remains HE-AAC.
+        let formatDescription = try audioFormatDescription(formatID: formatID)
+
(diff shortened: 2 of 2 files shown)


Sources/MUXSDKStatsInternal/Features/AudioTrackChangeEvents.swift
CFKevinRef, line 209: Is there a test for this method, and/or could we add one? It should be possible to manually create some format descriptions and validate the output.

CFKevinRef (changes requested):
Neat, makes sense to me. We should add a test though.

Sources/MUXSDKStatsInternal/Features/AudioTrackChangeEvents.swift
eugene-mux: Added a test for `codecString(for: CMFormatDescription)` using manually-created audio format descriptions. It covers the overload's format-ID → RFC 6381 mapping and the `mediaSubType` fallback.

One caveat: it does not reproduce the original HE-AACv2 regression, because manually-created `CMAudioFormatDescription` keep `mediaSubType` aligned with the `mFormatID` we pass in. The bug only appears on AVFoundation-demuxed assets, where `mediaSubType` is the compatible AAC-LC subtype while `CMAudioFormatDescriptionGetRichestDecodableFormat(...)` exposes the richer HE-AACv2 format - which can't be reconstructed synthetically. I verified that path manually with local HE-AACv2 playback (dashboard screenshot above).

Tests/MUXSDKStatsInternalTests/AudioTrackChangeEventsTests.swift
```diff
@@ -73,6 +73,27 @@ struct AudioTrackChangeEventsTests {
...
+        (AudioFormatID(kAudioFormatAC3), "ac-3"),
+    ])
+    func codecStringForFormatDescription(formatID: AudioFormatID, expected: String?) throws {
+        guard #available(iOS 15, tvOS 15, *) else {
+            return
+        }
+
+        // Synthetic format descriptions validate this overload and its fallback behavior.
+        // AVFoundation-created HE-AAC descriptions are still needed to exercise mediaSubType
+        // flattening to AAC-LC while the richest decodable format remains HE-AAC.
```
CFKevinRef, line 91: edit: I see the other comment now, this makes sense.

What's this comment referring to? It seems to suggest that AVFoundation will create format descriptions which will behave differently, or at least that `codecString(for:)` could behave differently with them.

CFKevinRef (approved):
tests look good too 👍🏻

Tests/MUXSDKStatsInternalTests/AudioTrackChangeEventsTests.swift
CFKevinRef: Ah I see what's going on here - an audio format description can contain a list of formats (see `CMAudioFormatDescriptionGetFormatList`) and `CMAudioFormatDescriptionGetRichestDecodableFormat`/`CMAudioFormatDescriptionGetRichestDecodableFormat` read from that list. This is coming from the magic cookie which does have a param in `CMAudioFormatDescriptionCreate`.

We could probably grab the magic cookie from the debugger on that manual case you used for verification, maybe even just try to reproduce the whole format description.

PR #370, opened 2026-07-14
update minimum deployment for examples and tests

CFKevinRef:
Xcode 27 won't even _build_ when deployment targets are set below iOS 15, and in practice these examples were not tested for such ancient OS versions anyway. Also these project files were fairly inconsistent. This PR sets a consistent 15.0 deployment target at the project level, removing target-specific overrides.

CocoaPods builds still use 12.0 in their project, but we can just remove CocoaPods support when we switch CI to Xcode 27 in the fall.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to example/test Xcode project settings and unit test assertions; shipped SDK deployment and runtime behavior are unchanged in this diff.
> 
> **Overview**
> Raises **minimum deployment targets** for example and fixture Xcode projects so they build on newer Xcode (e.g. below iOS 15 is rejected). **iOS** moves from 13/14 to **15.0** at the **project** level; **tvOS** goes from 13 to **15.0** in integration and unit test fixtures.
> 
> **Per-target** `IPHONEOS_DEPLOYMENT_TARGET` overrides are removed so app, test, and UI targets inherit the project default. Integration **ObjCTarget** drops an **18.4** iOS override; unit test targets drop redundant **Mac Catalyst / visionOS** deployment settings and empty `MACOSX_DEPLOYMENT_TARGET`.
> 
> In **`testPlaybackModeChange`**, test JSON is serialized with **`NSJSONWritingSortedKeys`** so playback-mode extra JSON expectations are stable across platforms; Mac Catalyst–specific expected strings and `#if` branches are removed in favor of **`XCTAssertEqualObjects`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35773f32df2ec6d2fb6f411011de38b26ba37ed2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


eugene-mux (approved):
Love it! Let's do the same for other SDKs as well.

PR #372, opened 2026-07-17
NAT-474: Simplify PlayerMonitor concurrency

CFKevinRef:
Effectively this commits to Combine, fully scheduled on the main queue, with a cancellation flag.

- Creates a single merged stream of events using Combine's built-in functionality. This replaces the ad-hoc method that sent events from various streams to a single subject. Now there is only one Combine pipeline.
- Forces all subscription operations to happen on the main queue. It seems likely that the [NAT-474] crash comes from subscriptions being torn down on background threads while an `onEvent` callback is actively executing on the main queue
- Adds a simple best-effort cancellation check before `onEvent` calls. Pipeline cancellation (via the `AnyCancellation` object created at initialization time) now has to queue up with everything else on the main queue, but this will short-circuit any further event callbacks.

Nice things like Atomics and OSAllocatedUnfairLock aren't available in iOS/tvOS 15, so it's a bit uglier than I'd prefer.

Resolves [NAT-474].

[NAT-474]: https://mux1.atlassian.net/browse/NAT-474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NAT-474]: https://mux1.atlassian.net/browse/NAT-474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes core player event observation and thread-safety around teardown, which could affect crash fixes (NAT-474) but also alter timing of rendition/bandwidth callbacks; minimum Swift monitor OS moves from 13 to 15.
> 
> **Overview**
> Refactors **`PlayerMonitor`** to use one merged Combine pipeline (rendition, text/audio track, and optional bandwidth events) instead of fanning multiple streams into a shared subject, and raises the Swift monitor’s minimum OS to **iOS/tvOS 15**.
> 
> Subscription setup and delivery are pinned to the main queue via **`ImmediateIfOnMainQueueScheduler`**, with a lock-backed **`SubscriptionWrapper`** so **`cancel()`** can tear down the **`AnyCancellable`** safely from any thread and skip **`onEvent`** after cancellation. **`MuxCore`** is imported with **`@preconcurrency`**, and the unsafe sendable workaround for events is removed.
> 
> In **`MUXSDKPlayerBinding`**, **`attachAVPlayer:`** assigns **`_player`** and the weak self reference before creating the Swift monitor so teardown and the periodic time observer share the same lifetime ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc3099e71eaaeb4d1d3848d883229926c57b064f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
+++ b/Sources/MUXSDKStats/MUXSDKPlayerBinding.m
@@ -154,17 +154,16 @@ static NSString *const RemoveObserverExceptionName = @"NSRangeException";
         NSLog(@"MUXSDK-ERROR - Cannot attach to NULL AVPlayer for player name: %@", _name);
         return;
     }
+    _player = player;
+    __weak typeof(self) weakSelf = self;
     if (@available(iOS 15, tvOS 15, *)) {
         self.shouldTrackRenditionChanges = NO;
-        __weak typeof(self) weakSelf = self;
         self.swiftMonitor = [[MUXSDKPlayerMonitor alloc] initWithPlayer:player onEvent:^(MUXSDKBaseEvent *event) {
             [weakSelf dispatchSwiftMonitorEvent:event];
         }];
     } else {
         self.shouldTrackRenditionChanges = YES;
     }
-    _player = player;
-    __weak MUXSDKPlayerBinding *weakSelf = self;
     _lastTimeUpdate = CFAbsoluteTimeGetCurrent() - MUXSDKMaxSecsBetweenTimeUpdate;
     _timeObserver = [_player addPeriodicTimeObserverForInterval:[self getTimeObserverInternal]
                                                           queue:NULL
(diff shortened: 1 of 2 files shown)


daytime-em:
> Probably worth breaking out the lock, flag, and subscription into a private inner class if we go with this pattern.

💯

daytime-em (approved):
This seems like a much safer approach! So much of the pre-Swift code assumes it's on the main queue, scheduling subscription and cancellation there makes perfect sense to me

I also agree that it would be smart to move the cancellation flag/lock/subscription into a private class, to protect `cancellationFlag`.

👍 , assuming that you'll probably hide the lock and stuff as mentioned

daytime-em (approved):
Greencheck reaffirmed :)

PR #375, opened 2026-09-01
NAT 494: Avoid missed/reordered events

CFKevinRef:
Fixes a class of bugs with the asynchronous Combine pipelines. At a high level, these pipelines lacked a buffer somewhere. Some asynchronous steps were effectively concurrent, with events reordered by which ones finished processing first. Other pipelines executed the asynchronous step one at a time, but this caused any incoming events to drop.

A small buffer fixes these issues, allowing push-based sources like notifications/observations to pool a few recent values while processing occurs.

Resolves [NAT-494]

[NAT-494]: https://mux1.atlassian.net/browse/NAT-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes event ordering and delivery for playback analytics pipelines; incorrect buffering could still drop or delay track/rendition events, but scope is limited to stats emission rather than playback control.
> 
> **Overview**
> Fixes **missed and out-of-order** MUX stats events when audio, text, and video rendition pipelines process slow async work (MainActor track loads, `Future`s) while notifications keep firing.
> 
> **Audio and text track change publishers** stop using `.switchToLatest()` on `mediaSelectionDidChange` (which dropped in-flight captures), emit the initial snapshot via `Just(...)`, and after debounce add a **size-5 buffer** with **serial** `flatMap(maxPublishers: .max(1))` so a few pending values can wait instead of being lost. Text track mapping switches to **`mapSerial`** for the `@MainActor` async step.
> 
> **Video rendition** (`timedRenditionInfoPublisher`) maps track changes to `Future`s without subscribing concurrently, then applies the same **buffer + serial flatMap** pattern before deduping on populated `MUXSDKVideoData`.
> 
> The shared **`Publisher.map` async helper is renamed to `mapSerial`** (tests updated) to reflect serial, one-at-a-time async transforms.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99ec254d84d1f7cd6fdd564f9aaadfae60c47dd9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/MUXSDKStatsInternal/Publishers+AVPlayerItem.swift
+++ b/Sources/MUXSDKStatsInternal/Publishers+AVPlayerItem.swift
@@ -57,7 +57,7 @@ extension AVPlayerItem {
         // playback, during seeking for example. While the following will unique these
         // objects, a new object here may still refer to the same rendition:
             .removeDuplicates()
-            .flatMap { videoAssetTrack in
+            .map { videoAssetTrack in
                 // Boost priority as this kicks off a chain of timing-sensitive operations
                 return Future(priority: .userInitiated) {
                     async let timing = self.currentTiming()
@@ -67,6 +67,8 @@ extension AVPlayerItem {
                     return await (timing, MUXSDKVideoData.makeWithRenditionInfo(track: videoAssetTrack, on: self))
                 }
             }
+            .buffer(size: 5, prefetch: .keepFull, whenFull: .dropOldest)
+            .flatMap(maxPublishers: .max(1)) { $0 }
         // Determine uniqueness (and therefore rendition changes) based on fully populated
         // video data:
             .removeDuplicates { timedInfoA, timedInfoB in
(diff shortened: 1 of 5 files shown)


Sources/MUXSDKStatsInternal/Publishers+AVPlayerItem.swift
```diff
@@ -57,7 +57,7 @@ extension AVPlayerItem {
         // playback, during seeking for example. While the following will unique these
         // objects, a new object here may still refer to the same rendition:
             .removeDuplicates()
-            .flatMap { videoAssetTrack in
```
CFKevinRef, line 60: `flatMap` has a default parameter of `maxPublishers: .unlimited`, which made this a concurrent _unordered_ asynchronous map.

daytime-em (approved):

